PR #94, opened 2021-04-02
WIP: Operation Migration plugin

dotansimha:
The goal of this plugin is to allow to migrate **operation** SDL + result based on pre-defined / custom rules. The goal is to allow schemas to evolve and update, without the need for deprecating fields that eventually bloats the schema, by allowing developers to change the running schema, but apply changes to the operation SDL/AST, variables JSON, and execution result. 

This is useful in cases where you are not in control of the executed operations (like, a mobile app that queries the server and can run old versions with old operations SDL). 

For example (migrate enum value that was changed):

```graphql
// OLD ENUM
enum Foo {
  A
  B
  OLD_VAL
}

// NEW ENUM
enum Foo {
   A 
   B
   NEW_VAL # this was changed from OLD_VAL
}
```

```ts
// Plugin usage:
useOperationMigration({
    migrations: [migrateEnum({ enumName: 'Foo', fromName: 'OLD_VAL', toName: 'NEW_VAL' })],
}),
```

Ideas / TODO:
- [x] Map all relevant places for migration
- [x] Initial implementation
- [x] Migrate enum value
- [x] Migrate field name
- [ ] Migrate field argument name
- [ ] Migrate scalar?
- [ ] Migrate type name? (mainly for fragments?)

### DIFF
--- a/packages/core/src/orchestrator.ts
+++ b/packages/core/src/orchestrator.ts
@@ -263,12 +263,18 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
       setResult: (newResult: AsyncIterableIterator<ExecutionResult> | ExecutionResult) => void;
     }) => void)[] = [];
     let context = args.contextValue || {};
+    let variables = args.variableValues || {};
 
     for (const onSubscribe of beforeCallbacks.subscribe) {
       const after = onSubscribe({
         subscribeFn,
         setSubscribeFn: newSubscribeFn => {
           subscribeFn = newSubscribeFn;
+        },
+        setVariables: newVariables => {
+          if (newVariables) {
+            variables = newVariables;
+          }
         },
         extendContext: extension => {
           context = { ...context, ...extension };
@@ -293,6 +299,7 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
     let result = await subscribeFn({
       ...args,
       contextValue: context,
+      variableValues: variables,
     });
 
     for (const afterCb of afterCalls) {
@@ -339,6 +346,7 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
         const afterCalls: ((options: { result: ExecutionResult; setResult: (newResult: ExecutionResult) => void }) => void)[] =
           [];
         let context = args.contextValue || {};
+        let variables = args.variableValues || {};
 
         for (const onExecute of beforeCallbacks.execute) {
           let stopCalled = false;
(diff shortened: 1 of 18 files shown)


packages/core/src/orchestrator.ts
```diff
@@ -263,13 +263,19 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
...
     let context = args.contextValue || {};
+    let variables = args.variableValues || {};
 
     for (const onSubscribe of beforeCallbacks.subscribe) {
       const after = onSubscribe({
         subscribeFn,
         setSubscribeFn: newSubscribeFn => {
           subscribeFn = newSubscribeFn;
         },
+        setVariables: newVariables => {
```
santino, line 274: Not related to this PR, but would make sense to introduce an `extendVariables` plugin that receives the context as well as current variables, and can merge new variables with passed one, or override variables.
Might be useful to implement logic to create variables given values from request headers.

packages/core/src/orchestrator.ts
```diff
@@ -263,13 +263,19 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
...
+          if (newVariables) {
+            variables = newVariables;
```
santino, line 276: do you want to override or merge variables?

packages/core/src/orchestrator.ts
```diff
@@ -348,6 +356,11 @@ export function createEnvelopOrchestrator(plugins: Plugin[]): EnvelopOrchestrato
             setExecuteFn: newExecuteFn => {
               executeFn = newExecuteFn;
             },
+            setVariables: newVariables => {
+              if (newVariables) {
+                variables = newVariables;
```
santino, line 361: same question here on overriding vs merging variables

packages/plugins/operation-migration/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "typings": "dist/index.d.ts",
+  "typescript": {
+    "definition": "dist/index.d.ts"
+  },
+  "scripts": {
+    "test": "jest",
+    "prepack": "bob prepack"
+  },
+  "dependencies": {
+    "@envelop/types": "0.2.1"
```
santino, line 33: `@graphql-tools/utils` is missing as a depencency here

santino (commented):
Feels pretty much like a rename transform.
Given we use the words "transforms" and "rename" in GraphQL Tools and GraphQL Mesh (just to name a couple), I wonder it this should renamed to something like "useTransformRename"

packages/core/src/orchestrator.ts
dotansimha: Yeah, and I think it's related because the variables API is being added in this PR :)

packages/core/src/orchestrator.ts
dotansimha: Probably we want to offer both options

YassinEldeeb:
Hey @dotansimha, this looks pretty interesting! I wonder what the current state of it now? 👀

PR #134, opened 2021-04-26
`usePersistedOperations`

dotansimha:
```ts
// This can be loaded from a remote source, or from a local json file, or just hardcoded
const data = {
   'persisted_1': 'query { ... }',
   'persisted_2' 'query { ... }',
};

const myStore: PersistedOperationsStore = {
  canHandle: idOrSdl => idOrSdl.startsWith('persisted_'),
  get: id => data[id] || null,
}

const envelopFn = envelop({
  plugins: [usePersistedOperations({ store })]
});
```

### DIFF
--- a/packages/plugins/persisted-operations/src/index.ts
+++ b/packages/plugins/persisted-operations/src/index.ts
@@ -1,0 +1,34 @@
+import { Plugin } from '@envelop/types';
+import { DocumentNode, parse } from 'graphql';
+
+export class NonPersistedOperationError extends Error {}
+
+export interface PersistedOperationsStore {
+  /**
+   * Check if the provided string content can be handled by the store or not.
+   * @param idOrSdl persisted id/hash, or SDL, provided for `parse`
+   */
+  canHandle(idOrSdl: string): boolean;
+  /**
+   * Transfroms a hash/id into a DocumentNode or SDL string. Returns `null` in case of a store miss.
+   * @param id
+   */
+  get(id: string): string | DocumentNode | null;
+}
+
+export const usePersistedOperations = (options: { store: PersistedOperationsStore }): Plugin => {
+  return {
+    onParse({ params, setParsedDocument }) {
+      if (typeof params.source === 'string' && options.store.canHandle(params.source)) {
+        const rawResult = options.store.get(params.source);
+
+        if (rawResult) {
+          const result = typeof rawResult === 'string' ? parse(rawResult) : rawResult;
+          setParsedDocument(result);
+        } else {
+          throw new NonPersistedOperationError(`The operation hash "${params.source}" is not valid`);
+        }
+      }
+    },
+  };
+};
--- a/packages/plugins/persisted-operations/src/index.ts
+++ b/packages/plugins/persisted-operations/src/index.ts
@@ -1,5 +1,5 @@
 import { Plugin } from '@envelop/types';
-import { DocumentNode, parse } from 'graphql';
+import { DocumentNode, GraphQLError, parse } from 'graphql';
 
 export class NonPersistedOperationError extends Error {}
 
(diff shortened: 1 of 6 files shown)


packages/plugins/persisted-operations/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Check if the provided string content can be handled by the store or not.
+   * @param idOrSdl persisted id/hash, or SDL, provided for `parse`
+   */
+  canHandle(idOrSdl: string): boolean;
+  /**
+   * Transfroms a hash/id into a DocumentNode or SDL string. Returns `null` in case of a store miss.
+   * @param id
+   */
+  get(id: string): string | DocumentNode | null;
```
n1ru4l, line 18: This should also allow returning a Promise for async stores 😇

dotansimha: Yeah, but it also means that `parse` becomes async, so I wasn't sure if it's the right path to go at the moment, because helix and other pipeline executors will need to adjust to that as well

dotansimha: I think maybe store data needs to be fetched ahead of time?

n1ru4l: TBH I don't have a use-case right now. I could just imagine that if you have a lot lot lot of documents. You don't wanna have all of them in the memory? 

This could still be implemented later, just one thought I had 🤔

dotansimha: You don't have to, it can be:

```
class MyStore implements PersistedOperationsStore {
  private data = {};
  
  load() { ... } // async, updates data 
  canHandle() { ... } // sync
  get() { ... } // sync
}

const store = new MyStore();
await store.load();

const envelopFn = envelop({
  plugins: [usePersistedOperations({ store })]
});
```

and if it's remote, it can even be triggered with a timeout to update the local store

n1ru4l: graphql-ws allows async persisted queries

```ts
useServer(
  {
    onSubscribe: async (_ctx, msg) => {
      const query = await queriesStore.load(msg.payload.query);
      if (!query) {
        // for extra security you only allow the queries from the store
        throw new Error('404: Query Not Found');
      }
      return {
        ...query,
        variableValues: msg.payload.variables, // use the variables from the client
      };
    },
  },
  wsServer,
);
```

dotansimha: we can try to add that, it just needs some adjustments to the core of `envelop`, because then the main function is async and the pipeline executors need to adjust to that.

PabloSzx:
Created a small PR that allows graphql-helix to accept an async parse.

https://github.com/contrawork/graphql-helix/pull/31

Envelop could only return a promise if the resulting value is a promise, and not be always an async function if it's not actually needed.

UPDATE: PR Merged and published

dotansimha:
> Created a small PR that allows graphql-helix to accept an async parse.
> 
> [contrawork/graphql-helix#31](https://github.com/contrawork/graphql-helix/pull/31)
> 
> Envelop could only return a promise if the resulting value is a promise, and not be always an async function if it's not actually needed.
> 
> UPDATE: PR Merged and published

Thanks @PabloSzx !  ❤️ 

My only real concern here is the fact that if we change `parse` to be `async` and return a `Promise`, it might break compatibility with other pipeline executors. 
I think at this point in time, I would like to keep `parse` sync, and stick to the same signature exported by `graphql`. 

I think eventually, stores should be in charge of loading and refreshing the store if needed, in an async way, and provide a sync API for `envelop`. I know it might be limiting, but I'm not sure that I want to diverge from the original spec at the moment.

PabloSzx:
> > Created a small PR that allows graphql-helix to accept an async parse.
> > 
> > [contrawork/graphql-helix#31](https://github.com/contrawork/graphql-helix/pull/31)
> > 
> > Envelop could only return a promise if the resulting value is a promise, and not be always an async function if it's not actually needed.
> > 
> > UPDATE: PR Merged and published
> 
> Thanks @PabloSzx !  ❤️ 
> 
> My only real concern here is the fact that if we change `parse` to be `async` and return a `Promise`, it might break compatibility with other pipeline executors. 
> I think at this point in time, I would like to keep `parse` sync, and stick to the same signature exported by `graphql`. 
> 
> I think eventually, stores should be in charge of loading and refreshing the store if needed, in an async way, and provide a sync API for `envelop`. I know it might be limiting, but I'm not sure that I want to diverge from the original spec at the moment. 

But the function doesn't have to be async if it doesn't have to (with conditionals "instanceof Promise") , that's what I'm saying, and the plugin that makes the parse async can have any possible warning that your implementation has to support it, and graphql-helix and envelop/app support it

Ps: Another option that prevents unnecessary conditionals being called in run-time is that the plugin gives a flag to envelop that automatically tells envelop to handle async parse functions

PR #149, opened 2021-04-28
New plugin: `useExtendedValidation` and implementation of `@oneOf` directive

dotansimha:
Background for this plugin:

At the moment, the existing phases of the requests pipeline are consist of: parse, validate, contextBuilding, execute. 

Variables parsing happens on a separate flow (as part of the request pipeline), and it's being provided to `execute` externally. Input coercion happens during execution, as an internal step. The implication of this is the fact that validation rules only have access to the AST of the documents (and to the validation context in general), but not to variables or execution params.

Modern implementation often needs to access variables in order to provide a more complete validation phase. 

To work around that, we can implement in `envelop` a plugin for extended validation. We follow the same `ValidationRule` interface, with the addition of `ExecutionArgs` provided. And Envelop makes sure to pass run that before `execute` runs.
The flow is the same as `validate` and it uses the same visitors flow, but allows developers to write more simplified code, and do more enriched validations.

As a simple example for this POC, I implemented the basics of this plugin and also implemented `@oneOf` directive for adding support for input unions.

More ideas that we might be able to implement:
* Allow having arguments/directives for scalars at the level of the usage, something like `field: String @max(5)`.
* Allow having validation rules that are aware of the user:  `field: String! @authorization(role: "test")`.

This might also introduce another way of doing directive-based validations, since it does not require composing resolvers.


packages/plugins/extended-validation/src/rules/one-of.ts
```diff
@@ -0,0 +1,27 @@
...
+        const argType = context.getArgument().type as GraphQLInputObjectType;
+        const directive = getDirectiveFromType(argType, 'oneOf');
+
+        if (directive) {
+          const variableName = node.value.name.value;
+          const variableValue = (executionArgs.variableValues || {})[variableName];
+          const keys = Object.keys(variableValue);
+
+          if (keys.length !== 1) {
+            throw new Error(`Exactly one key must be specified for argument of type ${argType} (used in $${variableName})`);
```
n1ru4l: Should this be more aligned with the API of a validation rule where you call `context.reportError`?

dotansimha: Yeah, you are right. I'll update that soon. For now it's still a draft because it needs to support a few more use-cases, and some checks are missing :)

PR #167, opened 2021-05-06
pre-iterate plugins

PabloSzx:
## Description

This a micro-optimization that works by assuming that plugins don't change (and they shouldn't), and it becomes a larger improvement when dealing with greater amount of plugins

## Type of change

**Test Environment**:

- OS: Windows 10 & macOS v11.3
- `@envelop/core`: v0.1.4
- NodeJS: v14.16.1 & v16.0.0

## Checklist:

- [ ] I have commented on my code, particularly in hard-to-understand areas

### DIFF
--- a/packages/core/src/create.ts
+++ b/packages/core/src/create.ts
@@ -29,6 +29,20 @@ export function envelop(options: { plugins: Plugin[]; extends?: Envelop[]; initi
   const childPlugins = (options.extends || []).reduce((prev, child) => [...prev, ...child._plugins], []);
   const plugins: Plugin[] = [...childPlugins, ...options.plugins];
 
+  const onContextBuildingCbs: NonNullable<Plugin['onContextBuilding']>[] = [];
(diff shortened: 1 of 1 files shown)


dotansimha:
This is great. Thank you @PabloSzx . 
I was thinking about wrapping callbacks functions in a wrapping API, so it will be something like that:

```
const afterCb = hooks.onParse(parseCbs);

// ... 

afterCb();
```

This way the code will be a bit more clean and without the `for` loops in the main file - what do you think?

PabloSzx:
> This is great. Thank you @PabloSzx .
> I was thinking about wrapping callbacks functions in a wrapping API, so it will be something like that:
> 
> 
> ```
> const afterCb = hooks.onParse(parseCbs);
> 
> // ... 
> 
> afterCb();
> ```
> 
> This way the code will be a bit more clean and without the `for` loops in the main file - what do you think?

Sorry for not replying that suggestion right away, the issue with that is that callbacks inside the loops use some variables in it's scope, it would need a bigger refactor of everything, unless I'm misunderstanding the idea

PR #171, opened 2021-05-09
fix graphql-jit implementation with lru cache

PabloSzx:
## Description

The current graphql-jit implementation doesn't work as intended since the result of the compilation is meant to be cached, and not being compiled on every execution, as it was detected while using it in benchmarks that using the plugin meant worse performance than not using it.

## Type of change

**Test Environment**:

- OS: macOS v11.3
- `@envelop/core`: v0.1.14
- NodeJS: v16.0.0

## Checklist:

### DIFF
--- a/packages/plugins/graphql-jit/src/index.ts
+++ b/packages/plugins/graphql-jit/src/index.ts
@@ -1,18 +1,81 @@
 /* eslint-disable no-console */
 import { Plugin } from '@envelop/types';
-import { GraphQLError } from 'graphql';
-import { compileQuery, isCompiledQuery, CompilerOptions } from 'graphql-jit';
+import { GraphQLError, DocumentNode, ExecutionResult, Source } from 'graphql';
+import { compileQuery, isCompiledQuery, CompilerOptions, CompiledQuery } from 'graphql-jit';
+import lru from 'tiny-lru';
+
+const DEFAULT_MAX = 1000;
+const DEFAULT_TTL = 3600000;
+
+export type JitCacheOptions = {
+  /**
+   * Maximum size of LRU Cache
+   * @default 1000
+   */
+  max?: number;
+  /**
+   * TTL in milliseconds
+   * @default 3600000
+   */
+  ttl?: number;
+};
 
 export const useGraphQlJit = (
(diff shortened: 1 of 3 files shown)


packages/plugins/graphql-jit/src/index.ts
```diff
@@ -1,18 +1,81 @@
...
+};
 
 export const useGraphQlJit = (
   compilerOptions: Partial<CompilerOptions> = {},
-  pluginOptions: Partial<{
-    onError: (r: ReturnType<typeof compileQuery>) => void;
-  }> = {}
+  pluginOptions: {
+    onError?: (r: ReturnType<typeof compileQuery>) => void;
+  } & JitCacheOptions = {}
```
PabloSzx: Here I was just following the existing style in other plugins, but I think it should be better to create and use interfaces of all the options and export them. But maybe the best solution for now would be to leave it as it is and then fix all these types alongside fixing #172

packages/plugins/graphql-jit/src/index.ts
```diff
@@ -1,18 +1,81 @@
...
 ): Plugin => {
+  const max = typeof pluginOptions.max === 'number' ? pluginOptions.max : DEFAULT_MAX;
+  const ttl = typeof pluginOptions.ttl === 'number' ? pluginOptions.ttl : DEFAULT_TTL;
+
+  const DocumentSourceMap = new WeakMap<DocumentNode, string>();
+
+  const jitCache = lru<
```
dotansimha: Why do we need to use `ExecutionResult`? Shouldn't it be the result of `compileQuery`?

ardatan: `compileQuery` returns ExecutionResult or CompiledQuery. I think it returns ExecutionResults if the query is not valid or something. See https://github.com/zalando-incubator/graphql-jit/blob/master/src/execution.ts#L190

PabloSzx: I think it's better to type it as `ReturnType<typeof compileQuery>`, the types look way cleaner and are more resilient to changes too, change just added

PR #179, opened 2021-05-11
feat(extended-validation): oneOf validation on input object type fields / array and nullable fields support

n1ru4l:
This is a follow up for https://github.com/dotansimha/envelop/pull/163

~~See https://github.com/n1ru4l/envelop/compare/n1ru4l:feat-one-of-extensions...n1ru4l:feat-one-of-input-field for the diff. This branch should be rebased after https://github.com/dotansimha/envelop/pull/163 is merged.~~

This implements support for `@oneOf` on nested fields. This is handy if you have complex filter masks (see https://api-docs.cyanite.ai/docs/graphql/inputs/experimental-similar-tracks-filter-bpm for an example).

~~It currently adds the `lodash.get` dependency, but I guess there is a smaller alternative somewhere that could be used instead.~~

It also changes the implementation to correctly handle lists of input types and nullable arguments/input fields that are of an input type.

### DIFF
--- a/packages/plugins/extended-validation/src/rules/one-of.ts
+++ b/packages/plugins/extended-validation/src/rules/one-of.ts
@@ -38,7 +38,7 @@ export const OneOfInputObjectsRule: ExtendedValidationRule = (validationContext,
           const argType = fieldType.args.find(typeArg => typeArg.name === arg.name.value);
 
           if (argType) {
-            traverseVariables(validationContext, arg, argType.type, [arg.name.value], values[arg.name.value]);
+            traverseVariables(validationContext, arg, argType.type, values[arg.name.value]);
           }
         }
       }
@@ -50,7 +50,6 @@ function traverseVariables(
   validationContext: ValidationContext,
   arg: ArgumentNode,
   graphqlType: GraphQLInputType,
-  path: Array<string | number>,
   currentValue: unknown
 ) {
   // if the current value is empty we don't need to traverse deeper
@@ -63,7 +62,7 @@ function traverseVariables(
     // if it is a list type currentValue MUST be an array
     assertArray(currentValue);
     currentValue.forEach((value, index) => {
-      traverseVariables(validationContext, arg, graphqlType.ofType, [...path, index], value);
+      traverseVariables(validationContext, arg, graphqlType.ofType, value);
     });
     return;
   }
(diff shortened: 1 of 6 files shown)


packages/plugins/extended-validation/src/rules/one-of.ts
```diff
@@ -38,22 +38,54 @@ export const OneOfInputObjectsRule: ExtendedValidationRule = (validationContext,
...
+    assertObject(currentValue);
+    if (Object.keys(currentValue).length !== 1) {
+      validationContext.reportError(
+        new GraphQLError(`Exactly one key must be specified for input type "${inputType.name}"`, [arg])
+      );
+    }
+  }
+  if (inputType instanceof GraphQLInputObjectType) {
+    // if it is an input type the argValue MUST be an object
+    assertObject(currentValue);
```
n1ru4l: I think we should not assert here and instead skip the validation if currentValue is not an object. The original validation should complain.

packages/plugins/extended-validation/src/rules/one-of.ts
```diff
@@ -38,22 +38,54 @@ export const OneOfInputObjectsRule: ExtendedValidationRule = (validationContext,
...
+) {
+  // if the current value is empty we don't need to traverse deeper
+  // if it shouldn't be empty, the "original" validation phase should complain.
+  if (currentValue == null) {
+    return;
+  }
+
+  if (isListType(graphqlType)) {
+    // if it is a list type currentValue MUST be an array
+    assertArray(currentValue);
```
n1ru4l: I think we should not assert here and instead skip the validation if currentValue is not an array. The original validation should complain.

packages/plugins/extended-validation/test/one-of.spec.ts
```diff
@@ -204,7 +371,84 @@ describe('oneOf', () => {
...
+        [
+          'Invalid: oneOf input list with incorrect list value (variant with weird behavior)',
+          {
+            document: `query user($input: [UserUniqueCondition!]) { listOneOf(input: $input) }`,
+            variables: {
+              input: { a: 1 }
+            },
+            expectedError: `Variable "$input" got invalid value { a: 1 }; Field "a" is not defined by type "UserUniqueCondition".`,
+          },
+        ],
```
n1ru4l: This one is a bit weird. Does it seem like graphql is automatically wrapping an object in an array?

dotansimha (approved):

PR #183, opened 2021-05-13
feat: hook into onExecuteSubscriptionEvent and useContextPerSubscriptionEvent

### LINKED ISSUE #80: Subscription: Create Context per published value

I have been thinking about how context and subscription behave after finding out that in almost every server implementation the context is retained for the whole subscription execution.

In a production app that somehow got slipped and lead to caching issues with DataLoaders, as each time the AsyncIterator of the subscription field published a new value the DataLoaders were hitting the cache (but the underlying data in the database had changed). The temporary solution was to disable the cache altogether and instead only use the DataLoaders for solving the n+1 database queries issue.

However, I don't think that this should be the way to go. A DataLoader cache should only be disabled in the context of a mutation operation as per spec definition a query and subscription operation do not change/mutate data and thus if a nested resolver tries to load the same data as a parent data using a cache instead of taking another roundtrip to the database server is perfectly fine.

Since a subscription can be long-running it makes sense to clear the cache after a value published by an AsyncIterator went through the resolver phase.

Currently, there is no straightforward way to do this and I have the feeling that from a UX experience it would be the simplest to just dispose of the context after the resolver phase and re-create it once a new value is published by the AsyncIteraror (vs. calling a lot of cache.clear() functions for each DataLoader or other caching mechanism).

One unsolved question is how that might work if the AsyncIterator event source is attached to the context. Usually, the source is passed to the `subscribe` handler via the context, but instead of having one source for each incoming request the source is often only initiated once upon server start and them simply forwarded by attaching it to the context (which is usually created for each incoming operation). So that wouldn't be an issue I guess.

Maybe, such behaviour could be a fit for `envelop` 😊

Currently, the only suboptimal solution would be to either disable caching completely (which could be expensive) or raise this question on the graphql-js reference implementation

**Additional context**

-

n1ru4l:
This is the first attempt of trying to solve https://github.com/dotansimha/envelop/issues/80

Example Usage:

```tsx
import { envelop } from "@envelop/core";

  plugins: [
    useContextPerSubscriptionEvent(async () => ({
      contextValue: { someValue: 1 },
      onEnd: () => cleanupResources()
    }))
  ]
});

const { subscribe } = getEnveloped();
```

There are several things that are currently not optimal

## 1. `subscribe` override

I created a copy of `subscribe` from `graphql-js` (which is surprisingly lightweight) and adjusted it so a custom `execute` function can be used for the `ExecuteSubscriptionEvent` phase.

I assume that the reason why it is not possible to provide a custom `execute` to the original `subscribe` from `graphql-js` is that the usage of `execute` is considered an implementation detail.

There is a pull request that attempts to allow overriding the `execute` passed to subscribe but no update from the `graphql-js` maintainers has been made recently. (https://github.com/graphql/graphql-js/pull/2485) I also created https://github.com/graphql/graphql-js/pull/3071

The plugin hard overrides the `subscribe` function. I cannot imagine any use-cases right now where you would want to compose `subscribe` functions (but I for example compose `execute` functions within graphql-live-query, so there might be a niche for that also with `subscribe`).

Ideally, I would like to avoid having a copy of `subscribe`, but I see no other option right now.

## 2. Context composition hook compatibility

This does currently not work with context composition, as we use a custom factory for creating the context. It would be interesting to get more insight on this from different people. I usually use the same context factory function for all operation types (mutation, subscription, or query). So maybe instead of having a custom `createContext` factory function, we should instead construct the context the same way as for `execute` (or `subscribe`). Probably the `useContextPerSubscriptionValue` `createContext` argument should be optional and instead, the implementation should use the default composition way of creating the context if the argument is not provided. Right now you would have to do sth like this:

```tsx
import { envelop } from "@envelop/core";

let contextFactor = null;
  plugins: [
    useContextPerSubscriptionEvent(async () => ({
      contextValue: await contextFactory()
    }))
  ]
});

const { subscribe, _contextFactory } = getEnveloped();
contextFactory = _contextFactory;
```


packages/core/src/subscribe.ts
```diff
@@ -0,0 +1,56 @@
+import { execute as defaultExecute, createSourceEventStream } from 'graphql';
+import isAsyncIterable from 'graphql/jsutils/isAsyncIterable';
+import mapAsyncIterator from 'graphql/subscription/mapAsyncIterator';
+import { makeSubscribe } from './util';
+
+/**
+ * This is a almost identical port from graphql-js subscribe.
+ * The only difference is that a custom `execute` function can be injected as an additional argument.
+ */
+export const subscribe = makeSubscribe(async args => {
```
dotansimha: Maybe this can live inside the plugin, and then you can do `setSubscribeFn` and replace the function?

dotansimha (changes requested):

packages/core/src/subscribe.ts
n1ru4l: Would prefer it to be a core or external plugin?

n1ru4l: Also, I am not sure how we should treat the subscribe function typings. Let's say we only use this custom subscribe function via a plugin. Should we still use the typings I defined in packages/types/src/index.ts for the subscribe function users can provide? How would the hooks API work for custom plugins that want to hook into the subscribe execution phase?

dotansimha: Not sure about the types, but you are correct, it needs to be adjusted somehow. The plugin for now can assume that it replaces the function and then call it with the custom params? Not 100% sure actually. 

I think `useContextPerSubscriptionEvent` should be a separate plugin package because it adds a specific functionality that isn't needed for most users. So we can have that as a separate plugin, and then we can have the custom subscribe method there (and you can use `setSubscribeFn` to override it)

n1ru4l: I removed the new hooks and instead moved stuff into a separate plugin that just calls `setSubscribeFn`. I actually think this plugin is needed by anyone using DataLoader and subscriptions.

Maybe we can move it back to the core once graphql-js allows providing a custom execute to subscribe. For now, all the types are aligned.

n1ru4l: As a follow up we maybe can also add a generic `subscribeExecute` that allows hooking into all phases of the execute call.

packages/plugins/execute-subscription-event/README.md
```diff
@@ -0,0 +1,23 @@
...
+Create a new context object per `ExecuteSubscriptionEvent` phase, allowing to bypass common issues with context objects such as [`DataLoader` caching issues](https://github.com/dotansimha/envelop/issues/80).
+
+```ts
+import { envelop } from '@envelop/core';
+import { useContextValuePerExecuteSubscriptionEvent } from '@envelop/execute-subscription-event';
+
+const getEnveloped = envelop({
+  plugins: [
+    useContextValuePerExecuteSubscriptionEvent(() => ({
+      contextValue: {
```
dotansimha, line 16: @n1ru4l maybe we should append this context to the general one? 
I think it might be an issue that we have separate context created here, and not having the original one as well, since resolvers might depend on the context.

packages/plugins/execute-subscription-event/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@envelop/execute-subscription-event",
+  "version": "0.1.0",
+  "author": "Dotan Simha <dotansimha@gmail.com>",
```
dotansimha, line 4: Update this ;)

packages/plugins/execute-subscription-event/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@envelop/execute-subscription-event",
+  "version": "0.1.0",
```
dotansimha, line 3: You can start with `0.0.0` and then changeset will bump it first.

packages/plugins/execute-subscription-event/src/subscribe.ts
```diff
@@ -0,0 +1,49 @@
...
+import isAsyncIterable from 'graphql/jsutils/isAsyncIterable';
+import mapAsyncIterator from 'graphql/subscription/mapAsyncIterator';
+import { makeSubscribe } from '@envelop/core';
+import { ExecuteFunction, SubscribeFunction } from '@envelop/types';
+
+/**
+ * This is a almost identical port from graphql-js subscribe.
+ * The only difference is that a custom `execute` function can be injected for customizing the behavior.
+ */
+export const subscribe = (execute: ExecuteFunction): SubscribeFunction =>
```
dotansimha, line 11: Some plugins might use `setExecuteFn` during `onExecute` with the assumption that they override this function.
Does it mean that these plugins need somehow to override this function here ?

dotansimha (changes requested):
Looking good. just some minor changes :) (and we need a rebase :P )

PR #187, opened 2021-05-14
additional defer/stream/subscription helix example

n1ru4l:
The helix example could act as a mirror for the experimental graphql playground repo 🤔

image


package.json
```diff
@@ -48,7 +48,8 @@
     "jest": "26.6.3"
   },
   "resolutions": {
-    "@changesets/git": "1.1.1"
+    "@changesets/git": "1.1.1",
+    "graphql": "15.4.0-experimental-stream-defer.1"
```
n1ru4l: This is a bit annoying but required in order to prevent the graphql.js mixed realm issue of having two versions installed at the same time.

Unfortunately, the stream-defer release line is not kept up2sync with main, there is currently no 15.5.x release available.

packages/core/src/create.ts
```diff
@@ -378,13 +378,20 @@ export function envelop({ plugins }: { plugins: Plugin[] }): Envelop {
...
+          for (const afterCb of afterCalls) {
+            afterCb({
+              result,
+              setResult: newResult => {
+                result = newResult;
+              },
+            });
+          }
+        } else {
+          // TODO: implement onExecuteDone invocation for async iterables
```
n1ru4l: Had to adjust this (as the compiler complained about the return type), also it is currently not clear how we should handle AsyncIterables returned from `execute`. See `onExecuteDone` section over here: https://github.com/dotansimha/envelop/issues/182

We should probably keep a copy of the execute and subscribe function types within envelop, so we can adjust them without relying on using experimental graphql-js versions. (I started doing something similar in #183 https://github.com/dotansimha/envelop/pull/183/files#diff-e71f2026e90f1ed0c00f8711eb855f0a8810f33dbf1491f5da9c97cdd513e5b4R22-R57)

package.json
dotansimha: Got it, but it means that the entire repo is using this experimental version, so maybe we can detach the example directory from the monorepo workspace?

n1ru4l: I rebase on main and detached the helix example from the main repo, another drawback is that we cannot simply link envelope core as it results in the graphql from two realm issue again, which is a bit of a bummer when you want to make changes in envelop and see how it affects the example. But I guess then you can temporarily add the resolutions entry for the desired graphql version and then use yarn link.

packages/core/src/create.ts
n1ru4l: I removed this for now.

package.json
dotansimha: Yeah, that's fine :) I think we can merge this one as-is, just to have it as a reference and we can later see if this could be improved (maybe with pnpm?)

dotansimha:
Oh @n1ru4l , just noticed that you applied that on top of the original helix example. I wanted to keep it clean and straightforward with just Query/Mutation. Can you please clone this dir into `examples/graphql-helix-stream-defer` directory?

n1ru4l:
@dotansimha Done!

maraisr:
Instead of fastify may I suggest we use polka in examples? It's faster than fastify 👀

dotansimha:
> Instead of fastify may I suggest we use polka in examples? It's faster than fastify 👀

Hi @maraisr ! We are always open for new ideas. In this case, we are trying to add examples for reference, so maybe we can have an example with `polka` in addition to fastify. I'm not highly familiar with polka - so feel free to create a PR ;)

PR #200, opened 2021-05-20
feat(core): useMaskedErrors plugin

n1ru4l:
Mask unexpected errors that occur during execution


packages/core/src/index.ts
```diff
@@ -1,8 +1,9 @@
...
-export * from './plugins/use-payload-formatter';
+export * from '@envelop/types';
+export * from './create';
+export * from './plugins/use-logger';
+export * from './plugins/use-timing';
+export * from './plugins/use-schema';
+export * from './plugins/use-error-handler';
+export * from './plugins/use-extend-context';
+export * from './plugins/use-payload-formatter';
+export * from './plugins/use-masked-errors';
```
PabloSzx: crlf commits?

n1ru4l: ahhh f windows :(

PabloSzx: if husky is added with pretty-quick, this issue won't ever happen 

@dotansimha can I create a PR?

n1ru4l: fixed 🕺

dotansimha (approved):

PR #201, opened 2021-05-20
add husky pre-commit pretty-quick

PabloSzx:


n1ru4l:
alternative title: stop me from committing clrf 🤣

PR #208, opened 2021-05-23
add husky pre-commit pretty-quick

PabloSzx:


package.json
```diff
@@ -48,7 +51,8 @@
     "jest": "26.6.3"
   },
   "resolutions": {
-    "@changesets/git": "1.1.1"
+    "@changesets/git": "1.1.1",
+    "prettier": "2.3.0"
```
dotansimha, line 55: Why do we need to downgrade the dependency here?

dotansimha (changes requested):

package.json
PabloSzx: that's not a downgrade, 2.3.0 is the latest, and it's because another dependency is installing an old 1.x version of prettier and it conflicts

dotansimha: Oops, I compared it to `3.1.0` of `pretty-quick` :D thanks @PabloSzx

dotansimha (approved):

PR #211, opened 2021-05-24
feat: execute/subscribe AsyncIterable API

n1ru4l:
~~This is a POC for hooking into the phases of the async iterable returned from execute or subscribe (similar to the approach described in https://github.com/dotansimha/envelop/issues/182).~~

It is currently based upon ~~https://github.com/dotansimha/envelop/pull/187 and~~ https://github.com/dotansimha/envelop/pull/183. It should be rebased once it is merged into main.~~

The idea is to let the `onSubscribeResult` and `onExecuteDone` handlers return an object with the properties `onNext` and `onFinal` which will be invoked in case we have an async iterable.

~~Currently, tests are missing, as I first want to gather feedback on whether this could be the way to go.~~

~~The relevant commit is https://github.com/dotansimha/envelop/pull/211/commits/aea90c4b054c461aab2cab48bdfdb6e67c9dfd33~~

### DIFF
--- a/packages/core/test/execute.spec.ts
+++ b/packages/core/test/execute.spec.ts
@@ -213,7 +213,7 @@ describe('execute', () => {
     await collectAsyncIteratorValues(result);
   });
 
-  it.skip('Should be able to invoke something after the stream has ended (manual return).', async () => {
+  it('Should be able to invoke something after the stream has ended (manual return).', async () => {
     expect.assertions(1);
     const streamExecuteFn = async function* () {
       for (const value of ['a', 'b', 'c', 'd']) {
@@ -235,7 +235,7 @@ describe('execute', () => {
                     latestResult = result;
                   },
                   onEnd: () => {
-                    expect(latestResult).toEqual(undefined);
+                    expect(latestResult).toEqual({ data: { alphabet: 'a' } });
                   },
                 };
               },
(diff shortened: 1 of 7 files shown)


packages/core/test/execute.spec.ts
```diff
@@ -133,4 +134,120 @@ describe('execute', () => {
...
+            };
+          },
+        },
+      ],
+      schema
+    );
+
+    const result: ReturnType<ExecuteFunction> = await teskit.executeRaw({} as any);
+    assertAsyncIterator(result);
+    result[Symbol.asyncIterator]().return!();
```
n1ru4l: AsyncIterables are so weird. When invoking return without ever a value published the finally block will never be invoked

packages/testing/src/index.ts
```diff
@@ -134,6 +134,13 @@ export function createTestkit(
 
       return r;
     },
+    executeRaw: makeExecute(async (args: ExecutionArgs) => {
+      const proxy = initRequest();
+      return await proxy.execute({
+        ...args,
+        contextValue: await proxy.contextFactory(args.contextValue),
+      });
+    }),
```
n1ru4l, line 143: @dotansimha At the moment the testkit is pretty helix tailored. I assume this is in order to be more integration test-like, and closer to how envelop is used with actual network layers?

I added this method here for now, but maybe we should in that case just construct all the stuff in place within the test?

dotansimha: I think we can remove helix for testing, I initially used it because I wanted to test envelop with a request, but it's no longer needed.

PR #223, opened 2021-05-26
Save id of persisted operation to the context

kamilkisiela:
No way to actually access context on parse phase so used a workaround.

### DIFF
--- a/.changeset/perfect-toys-compete.md
+++ b/.changeset/perfect-toys-compete.md
@@ -1,0 +1,5 @@
+---
+'@envelop/persisted-operations': minor
+---
+
+Save operation id to the context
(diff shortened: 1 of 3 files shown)


dotansimha (approved):

dotansimha:
> No way to actually access context on parse phase so used a workaround.

Maybe this is something we should change? Thoughts? @kamilkisiela @n1ru4l

kamilkisiela:
`onParse` could schedule a context modification that would be applied on the context building phase. I think this way it's nice and clean.

dotansimha:
@kamilkisiela you can now do it from `onParse` :)

kamilkisiela:
@dotansimha done

dotansimha (approved):

PR #245, opened 2021-06-02
feat(extended-validation): merge useExtendedValidation usages

n1ru4l:


dotansimha:
@n1ru4l now with the new `extendContext` that's available through all phases, maybe we can trick and collect all extended validations before `onExecute` and then just take it from the context and run it? :)

n1ru4l:
@dotansimha I did this in https://github.com/dotansimha/envelop/pull/245/commits/707164a1084418dd310925be2155980be82a0d25 the drawback here is that it now requires context to be non-null. Since extendContext expects a non-null context anyways maybe we can enforce it on a global level that the initial context cannot be a null/primitive value?

dotansimha:
> @dotansimha I did this in [707164a](https://github.com/dotansimha/envelop/commit/707164a1084418dd310925be2155980be82a0d25) the drawback here is that it now requires context to be non-null. Since extendContext expects a non-null context anyways maybe we can enforce it on a global level that the initial context cannot be a null/primitive value?

I think this is a fair assumption, since we make sure to set `{}` in case where context was not passed at all.

dotansimha (approved):

PR #259, opened 2021-06-08
ESM support for core and all plugins

PabloSzx:
## Description

ESM Support

## Type of change

## How Has This Been Tested?

**Test Environment**:
- OS: Windows & macOS
- NodeJS: v14.17.0 & v16.3.0

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas


dotansimha:
Thank you @PabloSzx ! 
Can you please help with some instructions for how to test this?

dotansimha (approved):

PabloSzx:
@dotansimha I can add an example in another PR that uses esm, but without an example is basically checking if neither the source code and .mjs files don't have any "require", that includes require(, require.resolve, require.cache etc

dotansimha:
> @dotansimha I can add an example in another PR that uses esm, but without an example is basically checking if neither the source code and .mjs files don't have any "require", that includes require(, require.resolve, require.cache etc

I see, thank you!

Maybe you can add as part of this PR, an example add under `/examples/with-esm` that uses the built ESM? Just for reference for developers, and I can use that as a base for testing.

PabloSzx:
> > @dotansimha I can add an example in another PR that uses esm, but without an example is basically checking if neither the source code and .mjs files don't have any "require", that includes require(, require.resolve, require.cache etc
> 
> I see, thank you!
> 
> Maybe you can add as part of this PR, an example add under `/examples/with-esm` that uses the built ESM? Just for reference for developers, and I can use that as a base for testing.

There is something tricky to have in mind, if we want the example to work as a real world example or an example as integration testing that imports every plugin to check if ESM doesn't break

dotansimha:
> > > @dotansimha I can add an example in another PR that uses esm, but without an example is basically checking if neither the source code and .mjs files don't have any "require", that includes require(, require.resolve, require.cache etc
> > 
> > 
> > I see, thank you!
> > Maybe you can add as part of this PR, an example add under `/examples/with-esm` that uses the built ESM? Just for reference for developers, and I can use that as a base for testing.
> 
> There is something tricky to have in mind, if we want the example to work as a real world example or an example as integration testing that imports every plugin to check if ESM doesn't break

If possible, I would like to see this example as a solution for both, what do you think? is it something we can have here? I mean, just HTTP+Helix+Envelop and ESM configured.

PabloSzx:
> If possible, I would like to see this example as a solution for both, what do you think? is it something we can have here? I mean, just HTTP+Helix+Envelop and ESM configured.

I don't think it makes sense for a real world example to import every plugin in the monorepo

PabloSzx:
and the other thing that applies to some of the other projects, just running an example with ESM doesn't ensure that everything works as expected, since the example would have to run everything since a require could be inside a function, a manual check of no "require"s in the source code and mjs files is required and I did in all the PRs I made, and the ones that are set as "draft pr" I'm waiting for a response of my comments since they will require some refactors that I would say are outside of the scope of just "ESM support"

Urigo:
@PabloSzx I think a single simple example of both ESM and CommonJS, not testing every plugin, for each repo, would give us enough confidence.
Later on if we want we can always increase the scope

examples/with-esm/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@envelop-examples/with-esm",
+  "version": "0.0.1",
+  "scripts": {
+    "dev": "tsup-node src/index.ts --format esm --watch src --onSuccess \"node dist/index.mjs\"",
+    "dev:cjs": "tsup-node src/index.ts --format cjs --watch src --onSuccess \"node dist/index.js\"",
+    "start": "tsup-node src/index.ts --format esm --onSuccess \"node dist/index.mjs\"",
```
PabloSzx: manually adding a `console.log` in the index.mjs file of @envelop/core you can see that the ESM support is being used correctly

dotansimha: Yeah that was exactly the missing piece for me :) Thanks @PabloSzx

dotansimha (approved):

PR #268, opened 2021-06-10
feat: add live query plugin

n1ru4l:


packages/plugins/live-query/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+      // @ts-expect-error: execute typings do not include AsyncIterable return right now
+      setExecuteFn(opts.liveQueryStore.makeExecute(executeFn));
+    },
+    onValidate: ({ addValidationRule }) => {
+      addValidationRule(NoLiveMixedWithDeferStreamRule);
+    },
+    onSchemaChange: ({ schema, replaceSchema }) => {
+      let patchedSchema = preparedSchemas.get(schema);
+      if (!patchedSchema) {
+        patchedSchema = addTypes(schema, [GraphQLLiveDirective]);
```
n1ru4l: Not sure whether adding the directive should be scope of the envelop tbh

dotansimha: I think we can expose it as string from the plugin package, so it will be easier to add it to the schema

dotansimha: And we can check in onSchemaChange if the schema has it or not

dotansimha (approved):

packages/plugins/live-query/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+```bash
+yarn add @envelop/live-query @n1ru4l/in-memory-live-query-store
+```
+
+## Usage Example
+
+```ts
+import { envelop, useSchema, useExtendContext } from '@envelop/core';
+import { createApplication } from 'graphql-modules';
```
dotansimha: unsued?

packages/plugins/live-query/README.md
```diff
@@ -0,0 +1,69 @@
...
+```tsx
+import { GraphQLSchema } from 'graphql';
+import { GraphQLLiveDirective } from '@envelop/live-query';
+
+const schema = new GraphQLSchema({
+  directives: [...specifiedDirectives, GraphQLLiveDirective],
+});
+```
+
+## Further information
```
dotansimha, line 84: Maybe add a short section about transports, and some links to examples ?

n1ru4l: done :)

PR #302, opened 2021-06-20
WIP: useGateway 

dotansimha:


packages/plugins/gateway/src/stitching.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    return fetchResult.json();
+  };
+
+  return {
+    loadSchema: async () => {
+      const schema = await introspectSchema(executor);
+
+      return makeRemoteExecutableSchema({
```
ardatan, line 46: This method is deprecated. We return the subschemaconfig directly to stitchSchemas to prevent multiple wrapping and reduce delegation rounds

PR #311, opened 2021-06-21
Smart Cache for Responses

kamilkisiela:
```typescript
interface Options<C = any> {
  /**
   * Maximum age in ms. Defaults to `Infinity`.
   */
  ttl?: number;
  /**
   * Overwrite the ttl for query operations whose execution result contains a specific object type.
   * Useful if the occurrence of an object time in the execution result should reduce the ttl of the query operation.
   */
  ttlPerType?: Record<string, number>;
  /**
   * Overwrite the ttl for query operations whose selection contains a specific schema coordinate (e.g. Query.users).
   * Useful if the selection of a specific field should reduce the TTL of the query operation.
   */
  ttlPerSchemaCoordinate?: Record<string, number>;
  /**
   * Allows to manually control the cache. Use `createController` to create a controller and pass it here.
   */
  controller?: Controller;
  /**
   * Allows to cache responses based on the resolved session id.
   * Return a unique value for each session.
   * Return `null` or `undefined` to mark the session as public/global.
   * Creates a global session by default.
   * @param context GraphQL Context
   */
  session?(context: C): string | undefined | null;
  /**
   * Skip caching of following the types.
   */
  ignoredTypes?: string[];
}
```


packages/plugins/response-cache/README.md
```diff
@@ -0,0 +1,13 @@
+## `@envelop/response-cache`
```
dotansimha, line 1: ADD DOCS

dotansimha: Also, add to list of plugins under website/lib/plugins.ts

packages/plugins/response-cache/src/index.ts
```diff
@@ -0,0 +1,208 @@
...
+              if (typeof id !== 'undefined') {
+                entitiesToRemove.add(makeId(typename, id));
+              }
+            });
+
+            entitiesToRemove.forEach(purge);
+          },
+        };
+      } else {
+        const operationId = createHash('md5').update(print(ctx.args.document)).digest('hex');
```
n1ru4l: how should caching per user session work?

n1ru4l: also sha might be better than md5

kamilkisiela: Based on a value defined in a context (custom function to set this up based on req.sessionId or whatever)

kamilkisiela: @n1ru4l I added the session thing

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+    await testInstance.execute(query);
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    await new Promise(resolve => setTimeout(resolve, 150));
+
+    await testInstance.execute(query);
+    expect(spy).toHaveBeenCalledTimes(2);
+  });
+
+  test('should cache responses based on session', async () => {
```
n1ru4l, line 488: 🥳

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+          }
+        }
+      }
+    `;
+
+    await testInstance.execute(query);
+    await testInstance.execute(query);
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    controller.purge('Comment', 2);
```
n1ru4l: That is pretty awesome, so you could do purge by listening to a database write log etc!

packages/plugins/response-cache/README.md
n1ru4l: https://github.com/dotansimha/envelop/pull/311/commits/c3b800b99f39e928a8917c7c2f25cc5965e1f1ec

packages/plugins/response-cache/src/in-memory-cache.ts
```diff
@@ -0,0 +1,101 @@
...
+      purgeResponse(responseId, false);
+    },
+  });
+
+  const entityToResponse = new Map<string, Set<string>>();
+  const responseToEntity = new Map<string, Set<string>>();
+
+  function purgeResponse(responseId: string, shouldRemove = true) {
+    // get entities related to the response
+    if (responseToEntity.has(responseId)) {
```
santino: When working with Maps I normally try to avoid using `has` and then `get` to reduce the matching operations to perform.
you could do `const cachedResponse = responseToEntity.get(responseId)` and then check if this has a value and reuse this var later on.
That would remove an opration.

santino (commented):
Just a small suggestion

packages/plugins/response-cache/src/in-memory-cache.ts
n1ru4l: https://github.com/dotansimha/envelop/pull/311/commits/b6ab03c1d4c4607138335463e55f0b147da11dd1

PR #318, opened 2021-06-22
useOperationFieldPermissions

### LINKED ISSUE #240: Operation validation based on schema access permissions

I need to limit schema access for different GraphQL API users.

The [`Type.attribute`](https://github.com/graphql/graphql-spec/blob/main/rfcs/SchemaCoordinates.md#typeattribute) syntax should be used for validating whether the viewer is allowed to execute an operation with a given selection set. In addition, it should be possible to use wildcards e.g. `Query.*` (allow selecting all fields on the `Query` type).

This is a perfect fit for useExtendedValidation, as it has access to the context form which we can read the user scope, which is a set of schema coordinates).  Developers can customize from where they want to load the list of allowed fields for each API user.

WIP name: `useOperationSelectionSetAuthorization`

Do the validation within the resolvers. But it is fiddly and does not stop the execution in the first place.

Quick prototype:

```tsx
import { envelop, Plugin, useEnvelop, useExtendContext } from "@envelop/core";
import {
  ExtendedValidationRule,
  useExtendedValidation,
} from "@envelop/extended-validation";
import { GraphQLError, TypeInfo, visitWithTypeInfo } from "graphql";
import { getWrappedType } from "./get-wrapped-type";

const getWildcardTypes = (scope: Set<string>): Set<string> => {
  const wildcardTypes = new Set<string>();
  for (const item of scope) {
    if (item.endsWith("*")) {
      const [typeName] = item.split(".");
      wildcardTypes.add(typeName);
    }
  }
  return wildcardTypes;
};

const operationScopeSymbol = Symbol();

const getSchemaCoordinatesFromContext = (input: unknown): Set<string> => {
  if (typeof input !== "object" || !input || !(operationScopeSymbol in input)) {
    throw new Error("OperationScopeRule was used without context.");
  }
  return input[operationScopeSymbol];
};

/**
 * Validate whether a user is allowed to execute a certain GraphQL operation.
 */
const OperationScopeRule: ExtendedValidationRule = (context, executionArgs) => {
  const typeInfo = new TypeInfo(context.getSchema());

  const allowedSchemaCoordinates = getSchemaCoordinatesFromContext(
    executionArgs.contextValue
  );
  const allowedWildcardTypes = getWildcardTypes(allowedSchemaCoordinates);
  return visitWithTypeInfo(typeInfo, {
    Field(node) {
      const parentType = typeInfo.getParentType();
      if (parentType) {
        const wrappedType = getWrappedType(parentType);
        const schemaCoordinate = `${wrappedType.name}.${node.name.value}`;

        if (
          !allowedWildcardTypes.has(wrappedType.name) &&
          !allowedSchemaCoordinates.has(schemaCoordinate) &&
          !allowedSchemaCoordinates.has("*") // everything!
        ) {
          context.reportError(
            new GraphQLError(
              `Insufficient permissions for selecting '${schemaCoordinate}'.`,
              [node]
            )
          );
        }
      }
    },
  });
};

type PromiseOrValue<TValue> = Promise<TValue> | TValue;

type OperationScopeOptions<TContext> = {
  getPermissionsFromContext: (context: TContext) => PromiseOrValue<Set<string>>;
};

export const useOperationSelectionSetAuthorization = <TContext>(
  opts: OperationScopeOptions<TContext>
): Plugin =>
  useEnvelop(
    envelop({
      plugins: [
        useExtendedValidation({
          rules: [OperationScopeRule],
        }),
        useExtendContext(async (context) => ({
          [operationScopeSymbol]: await opts.getPermissionsFromContext(
            context as TContext
          ),
        })),
      ],
    })
  );

```

n1ru4l:
This envelop plugin restricts operation execution via schema coordinates/type wildcards.

Closes https://github.com/dotansimha/envelop/issues/240

Feel free to propose a better plugin name

### DIFF
--- a/packages/plugins/operation-field-permissions/src/index.ts
+++ b/packages/plugins/operation-field-permissions/src/index.ts
@@ -1,4 +1,4 @@
-import { envelop, Plugin, useEnvelop, useExtendContext } from '@envelop/core';
+import { Plugin, useExtendContext } from '@envelop/core';
 import { ExtendedValidationRule, useExtendedValidation } from '@envelop/extended-validation';
 import { GraphQLType, GraphQLList, GraphQLNonNull, GraphQLError } from 'graphql';
 
@@ -76,17 +76,20 @@ type OperationScopeOptions<TContext> = {
 
 const defaultFormatError = (schemaCoordinate: string) => `Insufficient permissions for selecting '${schemaCoordinate}'.`;
 
-export const useOperationPermissions = <TContext>(opts: OperationScopeOptions<TContext>): Plugin =>
-  useEnvelop(
-    envelop({
-      plugins: [
+export const useOperationPermissions = <TContext>(opts: OperationScopeOptions<TContext>): Plugin => {
+  return {
+    onPluginInit({ addPlugin }) {
+      addPlugin(
         useExtendedValidation({
           rules: [
             OperationScopeRule({
               formatError: opts.formatError ?? defaultFormatError,
             }),
           ],
-        }),
+        })
+      );
+
+      addPlugin(
         useExtendContext(async context => {
           const permissions = await opts.getPermissions(context as TContext);
 
@@ -102,7 +105,8 @@ export const useOperationPermissions = <TContext>(opts: OperationScopeOptions<TC
           return {
             [OPERATION_PERMISSIONS_SYMBOL]: scopeContext,
           };
-        }),
-      ],
-    })
-  );
+        })
+      );
+    },
+  };
+};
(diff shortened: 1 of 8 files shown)


packages/plugins/operation-field-permissions/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  useEnvelop(
+    envelop({
+      plugins: [
+        useExtendedValidation({
+          rules: [
+            OperationScopeRule({
+              formatError: opts.formatError ?? defaultFormatError,
+            }),
+          ],
+        }),
```
n1ru4l: It made the most sense to re-use the existing plugin here 🤔

dotansimha (approved):

PR #324, opened 2021-06-23
Initial implementation of persisted queries plugin

santino:
## Description

This draft PR is for implementing a persisted queries plugin with the following goals:
- Handle multiple persisted queries list, to potentially support multiple clients integrating with a single GraphQL server
- Provide a basic implementation for a store that handle static JSON files (as generated by most common clients)
- Provide a mechanism to update the store at runtime, even targeting specific lists only (whitelist)

Since the plugin is meant to support multiple JSON files, it also offers an option to match from a single list only.
This allows developers to implement logic to target a single list, so you can avoid matching against all lists when you know your request comes from a specific application and so must be available in a single list.
The point is not necessarily to improve performance, in fact matching against multiple lists might still be implemented with o(1) cost (since it's just a lookup job), but more about offering proper support for multiple applications; in fact it might be that different applications have one or more equal queries, that would result in the same SHA256 id. By handling multiple and separate queries lists there is no worry of conflicts and file might just be provided to the server as they're generated by clients.

## Type of change

## How Has This Been Tested?

Being just a draft, this is open for a conversation.
I am planning to add tests once we finalise the strategy.

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

## Further comments

This plugin might be seen in conflict with the existing persisted-oeprations, but it differs as it is more intended (but not limited) to handle static lists.  
The main difference is the concept of handling multiple lists to support multiple client applications.  
Also this plugin provides an implementation for a store (again targeting a basic usage with static files), which is useful to be used as a guide for developers wanting to implement their own store from other sources (such as databases).

Ultimately I am open to merge the two plugins so that we offer a solution that is flexible enough and provides a basic store implementation.

### DIFF
--- a/packages/plugins/persisted-queries/README.md
+++ b/packages/plugins/persisted-queries/README.md
@@ -35,7 +35,7 @@ const getEnveloped = envelop({
 
 -----
 
-await persistedQueriesStore.build(); // get store ready, in this case by loading persisted-quries files
+await persistedQueriesStore.load(); // get store ready, in this case by loading persisted-quries files
 
 server.listen() // once queries are loaded you can safely start the server
 ```
--- a/packages/plugins/persisted-queries/README.md
+++ b/packages/plugins/persisted-queries/README.md
@@ -1,4 +1,4 @@
-## `@envelop/persisted-operations`
+## `@envelop/persisted-queries`
 
 TODO
 
@@ -9,8 +9,6 @@ yarn add @envelop/persisted-queries
 ```
 
 ## Basic Usage
-
-TODO
 
 For convenience the plugin expose a Class which allow you to setup a basic store from one or more JSON files.  
 You can use this or build your own store that implements `PersistedQueriesStore` typescript interface exported by the plugin.
@@ -29,28 +27,36 @@ const getEnveloped = envelop({
   plugins: [
     usePersistedQueries({
       store: persistedQueriesStore,
-      onlyPersisted: true,
-      setQueryId: (context) => context.request.body.queryId
+      onlyPersisted: true, // default `false`. When set to true, will reject queries that don't have a valid query id
     }),
     // ... other plugins ...
   ],
 });
 
-...
+-----
 
-const { parse, validate, contextFactory, execute, schema } = config.getEnveloped({ request });
-
-...
-
-await persistedQueriesStore.buildStore(); // load persisted-quries files
+await persistedQueriesStore.buildStore(); // get store ready, in this case by loading persisted-quries files
 
 server.listen() // once queries are loaded you can safely start the server
 ```
 
+TODO: explain default behaviour which identifies query id from standard "query" param (GraphQL source)
+
 ## Advanced usage
 
-TODO
+In order to build custom logic you probably want to pass the request object (or part of it) to `getEnveloped`, so that this will be available as the initial context
 
-- default behaviour to identify query id from standard "query" param
-- how to match ids from a single queries list
-- pass request object to getEnveloped, in order to set initial context necessary to access request data
+```ts
+httpServer.on('request', (request, response) => {
+  const { parse, validate, contextFactory, execute, schema } = getEnveloped({ request });
+  // ...
+}
+```
+
+TODO, describe the following:
+
+- lists are named after file name (without extension) when using `JsonFilesStore`
+- build custom logic to set query id: `setQueryId: (context) => context.request.body.queryId // set custom logic to get queryId`
+- how to match ids from a single queries list: `pickSingleList: (context) => `${context.request.headers.applicationName}_persistedQueries``
+- `setQueryId` and `pickSingleList` can return undefined, explain what happens
+- building your own store
--- a/packages/plugins/persisted-queries/README.md
+++ b/packages/plugins/persisted-queries/README.md
@@ -1,0 +1,51 @@
+## `@envelop/persisted-operations`
+
+TODO
+
+## Getting Started
+
+```
+yarn add @envelop/persisted-queries
+```
+
+## Basic Usage
+
+TODO
+
+For convenience the plugin expose a Class which allow you to setup a basic store from one or more JSON files.  
+You can use this or build your own store that implements `PersistedQueriesStore` typescript interface exported by the plugin.
+
+```ts
+import { resolve } from 'path';
+import { envelop } from '@envelop/core';
+import { usePersistedQueries, JsonFilesStore } from '@envelop/persisted-queries';
+
+const persistedQueriesStore = new JsonFilesStore([
+  resolve(process.cwd(), 'assets/clientOne_persistedQueries.json'),
+  resolve(process.cwd(), 'assets/clientTwo_persistedQueries.json'),
+]);
+
+const getEnveloped = envelop({
+  plugins: [
+    usePersistedQueries({
+      store: persistedQueriesStore,
+      onlyPersisted: true,
+      setQueryId: (context) => context.request.body.queryId
+    }),
+    // ... other plugins ...
+  ],
+});
+
+...
+
+await persistedQueriesStore.buildStore(); // load persisted-quries files
+server.listen() // once queries are loaded you can safely start the server
+```
+
+## Advanced usage
+
+TODO
+
+- default behaviour to identify query id from standard "query" param
+- how to match ids from a single queries list
+- 
--- a/packages/plugins/persisted-queries/README.md
+++ b/packages/plugins/persisted-queries/README.md
@@ -35,7 +35,7 @@ const getEnveloped = envelop({
 
 -----
 
-await persistedQueriesStore.buildStore(); // get store ready, in this case by loading persisted-quries files
+await persistedQueriesStore.build(); // get store ready, in this case by loading persisted-quries files
 
 server.listen() // once queries are loaded you can safely start the server
 ```
@@ -55,8 +55,9 @@ httpServer.on('request', (request, response) => {
 
 TODO, describe the following:
 
-- lists are named after file name (without extension) when using `JsonFilesStore`
+-  when using `JsonFilesStore`, lists are named after file name (without extension)
 - build custom logic to set query id: `setQueryId: (context) => context.request.body.queryId // set custom logic to get queryId`
-- how to match ids from a single queries list: `pickSingleList: (context) => `${context.request.headers.applicationName}_persistedQueries``
+- how to match ids from a single queries list: `` pickSingleList: (context) => `${context.request.headers.applicationName}_persistedQueries` ``
 - `setQueryId` and `pickSingleList` can return undefined, explain what happens
 - building your own store
+- update the store during runtime with updated or new lists, without restarting the server
--- a/packages/plugins/persisted-queries/README.md
+++ b/packages/plugins/persisted-queries/README.md
@@ -38,7 +38,12 @@ const getEnveloped = envelop({
 
 ...
 
+const { parse, validate, contextFactory, execute, schema } = config.getEnveloped({ request });
+
+...
+
 await persistedQueriesStore.buildStore(); // load persisted-quries files
+
 server.listen() // once queries are loaded you can safely start the server
 ```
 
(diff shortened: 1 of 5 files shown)


packages/plugins/persisted-queries/README.md
```diff
@@ -0,0 +1,63 @@
+## `@envelop/persisted-queries`
+
+TODO
+
+## Getting Started
+
+```
+yarn add @envelop/persisted-queries
```
n1ru4l, line 8: Isn't persisted-operations the better wording?

santino: Currently I am intentionally using this to differentiate from existing "persisted-operations" plugin.
This is currently a draft to discuss how to progress given overlap, if it encounters favour this might be merged within existing "persisted-operations"; it's something to discuss with project owners.

BTW, if you want my opinion, it's arguable.
You send a request to your server, and a single request can potentially contain multiple operations; yet it would have only a single hash/id.
In this case you're not persisting the operation (as you don't have an id per operation), but the whole query string body.
Hence, you are persisting the query string value, which content can involve multiple operations.

Also, persisted queries seems to be the de-facto convention, as documented in [Relay](https://relay.dev/docs/guides/persisted-queries/) and [Apollo](https://www.apollographql.com/docs/apollo-server/performance/apq/), just to mention some big players; but there are definitely more.

Personally, I would prefer to call it "persisted-documents" because in reality the value that you pass into query string is a "to-be parsed" GraphQL document, but I suppose that would be more difficult to guess, given it's not currently standardised like this.
In fact, if you look into the code, I am setting the hash as "documentId" within the context object; for potential later reference.

yaacovCR: https://github.com/graphql/graphql-spec/pull/777

Persisted documents seems the currently correct term.

documents, of course, are written in “graphQL” or even the graphql query language, still a little wierd

yaacovCR: Should we ride the edge?

santino: Sounds like I am backed, and so I would be happy to do so!
Will check with owners of the project and eventually change the name of this plugin to persisted-documents before fully opening the PR.

I think eventually, in case we decide to favour this over current "persisted-operations" plugin; the existing one could simply be deprecated and removed allowing this to become the standard.

Again, really looking forward to hear from The Guild on how to solve the overlap first.

tvvignesh: @santino Regarding the naming, we use the name "Persisted Operations" in Hive as of today. So, my view is that irrespective of what name we choose, we should try to stick to one name in both the plugin as well as hive to maintain consistency. Otherwise, both are explanatory according to me.

And in context with overlap with [persisted-operations](https://github.com/dotansimha/envelop/tree/main/packages/plugins/persisted-operations) - you are right, as a user it would sound very confusing to have both `persisted-operations` and `persisted-queries`

There is one question I have here when talking about the overlap, what are the differences you see from your perspective between the both plugins?

@kamilkisiela Thoughts?

santino:
Adding this separate comment to discuss more the differences with the existing "persist-operations" plugin.
I think the goals of the two are different. The existing plugin is highly customisable but relies on full implementation from developers, they can write their own store in any shape and build their matching logic.
This new plugin wants to be more of a Plug'n'Play solution, whilst also providing extra functionalities. This is possible by being more assertive on the shape of the store, in this case, I'm expecting a Map with a key/value pair where the key is the name of a list and the value is an object containing hashes and their respective stringified documents.

In all honesty, I wouldn't even be sure how to add the extra functionalities to the existing plugin without changing the existing API.
You cannot support multiple persisted lists if you don't know which shape your store has, at least not if you want to provide a ready-to-use matching logic; rather than just relying on developers building the whole thing on their own; which is a core goal of this plugin.
If you do change the API, then you introduce breaking changes and in the end, you do end up with a different plugin anyway, even if you keep it in the existing package and with the same name.

I could add a simple store implementation that handles static JSON files to the existing plugin, but then developers will still have to build their matching logic to retrieve the stringified documents from the store.
The above logic devs need to build, becomes more complex if they want to handle multiple lists. You could provide snippets in the README for doing this, but still, it remains code that developers are responsible for, they need to write it and maintain it, hence keeping the knowledge overhead; as such is not a Plug'n'Play solution.

My plugin provides a good solution for static JSON files, but this can be used as an example and customised for more complex setups (such as grabbing persisted lists from a database f.i.).
Yet, the only thing developers need to provide is just a Map with key/value pairs; it will work with a single list or multiple lists and they won't need to write nor maintain any logic to use their lists.

Hope this helps.

packages/plugins/persisted-queries/README.md
santino: My opinion is that persisted-documents is semantically correct and persisted-operations isn't, because as explained above you're not persisting operations but full GraphQL documents.
@yaacovCR seems to agree and has provided a good reference for glossary.

Having said this, as the owners of the project you are definitely free to make your decision, I just want to provide a recommendation.
I totally get your point that The Guild should use the same conventions across all their tools, it will surely be appreciated by your users :)

As about the question about the difference, I have added a separate comment to keep the conversations separate.

dotansimha: That's a endless discussion :D 
We used the term `documents` in codegen, and got some user rejections because they were used to `queries` (which doesn't make sense)... 

Basically, for me, it's:
* `document` - a DocumentNode, that can hold any kind of AST.
* `operation` - Anything executable (query/mutation/subscription).

packages/plugins/persisted-queries/src/plugin.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface PersistedQueriesStore {
+  get(): PersistedQueriesStoreList;
+}
+
+interface PluginContext {
+  [contextProperty]?: string;
+}
+
+export type UsePersistedQueriesOptions = {
```
dotansimha, line 17: I wonder if these can be moved into the existing `persisted-operations` plugin? 
And then `pickSingleList` might be flag for the store?

packages/plugins/persisted-queries/src/plugin.ts
```diff
@@ -0,0 +1,96 @@
...
+export const usePersistedQueries = (rawOptions: UsePersistedQueriesOptions): Plugin<PluginContext> => {
+  const options: UsePersistedQueriesOptions = {
+    ...DEFAULT_OPTIONS,
+    ...(rawOptions || {}),
+  };
+
+  return {
+    onParse({ context, params, extendContext, setParsedDocument }) {
+      const store = options.store.get(); // retrieve fresh instance of store Map
+      const queryId = options.setQueryId ? options.setQueryId(context) : queryIdFromSource(params.source);
```
dotansimha, line 37: I think we have a similar thing in persisted-operations plugin, that @kamilkisiela added

dotansimha (commented):
I like this concept. I wonder if some of the things can be merged with persisted-operations plugin? 🤔

packages/plugins/persisted-queries/README.md
santino: And there is also Source, to throw more stuff into the mix.
Technically the string containing the Document, which includes the Operations, is a GraphQL Source.

Anyway, was a nice discussion and I care a lot about semantics, but I don't feel like insisting.
Persisted-operations be it, if you feel like so :)

packages/plugins/persisted-queries/src/plugin.ts
santino: Yes, but then you need to be prescriptive of store shape (key/value Map).
So your signature changes.

packages/plugins/persisted-queries/src/plugin.ts
santino: I think currently existing persisted-operations doesn't receive context, to begin with.
Also there is no logic to grab the id/hash from standard GraphQL source.

Probably the bit you're referring to has been introduced by @kamilkisiela in Hive rather then Envelop

santino:
Some, probably, you could define an easy way to get the id from context and implement default behaviour to get it from GraphQL source.
You could also provide a simple store implementation for static files.

Anything else, like a built-in matching logic that works out-of-the-box with one/multiple lists, would require breaking changes in the plugin.

The best way I could think about implementing those functionalities is the way I built this new plugin.

So the question still remains.
Do you think we should have both plugins as part of Envelop?
If not, which one should we prioritise?
We introduce concepts implemented here onto existing persisted-operations, but at the condition of introducing breaking changes?
Or do we release a new plugin and deprecate the existing one?

dotansimha:
> Some, probably, you could define an easy way to get the id from context and implement default behaviour to get it from GraphQL source.

@kamilkisiela did `writeToContext` flag in the existing plugin, in order to make it compatible. We can also add a function to extract the id from the context, and then you can use `context.req` if it's there, and provide custom function for that. 

 

> You could also provide a simple store implementation for static files.

The way I see it, the store should have a single API for `get`, like we have today, and then behind that, you can implement any store, based on any amount of files. 



> Anything else, like a built-in matching logic that works out-of-the-box with one/multiple lists, would require breaking changes in the plugin.

I'm not sure that we have to break it, can you please explain why we can do a single store, that supports multiple files? eventually it doesn't matter which file the data is fetched from, it's just needs the id and the document. 





> Do you think we should have both plugins as part of Envelop?

I still think we can find a way to keep the persisted operations API simple as possible, and allow the store to implement any kind of flow. 

Maybe the API for ` get(id: string): string | DocumentNode | null;` can be ` get(id: string): { document: string | DocumentNode | null, customData: any };` and then each store can return custom data to be added to context? (so if it's based on multiple stores or files, and you need to know about that you can add it there).

santino:
> We can also add a function to extract the id from the context

yes, this can easily be added, and also a default behavior to get id from source, as I have implemented [here](https://github.com/dotansimha/envelop/pull/324/files#diff-d714d01391f4c5f4fa0b58e77b279784368e95177c6752098c93fbb31dc6e9c6R63) (which is effectively comparable to a default implementation for `canHandle` as currently expected by "persisted-operations" plugin.

>  can you please explain why we can do a single store, that supports multiple files

The main use case is to support multiple applications. Each application exports its list of persisted queries. Some applications might require one or more exact queries that will have conflicting hashes. The conflict is not necessarily a problem, cause you can ignore duplicate hashes, trusting that their value will always be equal.

But when I support multiple applications I might want to tell my server that the requests come from application _A_ and it should match queries only from the list generated by application _A_ rather than all applications supported by my server.
It might not necessarily be a performance concern, because after all, we're still talking about O(1) and not O(n); but it might be a security concern: I want clients to disclose who they're, and allow them to access only queries defined by their themselves.
Let's not forget that persisted queries are not just about improved performance, but also security improvement as you're not disclosing the schema openly.

> I'm not sure that we have to break it

With the context above, if you do want to support multiple lists, you need a signature for the store, which is no longer a simple key/value pair.

> I still think we can find a way to keep the persisted operations API simple as possible

This is my main goal, if you look at the README file you can see how you can simply pass one or more json files to the plugin and it works out of the box; the API is damn simple in my view.

If you do want to implement your own store, the only overhead I am proposing (to prescript the store shape) is to wrap each list in a Map which key is the name of the list and value is the full object containing hashes and their values.
It's literally a wrapping layer, that does not make the API more complex.
In spite of this wrapping layer, you do not need to pass `canHandle` nor `get` functions in the new API.
You can, for a more complex setup (that you are willing to add to the existing plugin) use `setQueryId` to retrieve query id in a custom way (e.g. from context).

> and allow the store to implement any kind of flow

Do you think the plugin I am proposing would prevent anything here?

As for your proposal to extend `get` API, yes, probably that could work.
But you would still rely on developers building their own logic to support multiple lists in a single store, within the get function; where my goal is precisely to support one and multiple lists out-of-the-box.

I am only addressing your points, but really do not feel like I am forcing you to accept this new plugin.
In the same way, I am addressing your points I would invite you to reflect, and maybe share, how my proposed plugin might affect them negatively, or offer an API that is not as simple as the existing one.
At the end of the day, I appreciate my plugin is different from the existing one, but if it does tick all the boxes, with simple API design, extended functionalities, and the ability to fain-grain custom usages; it might be good regardless.

dotansimha:
We are going with https://github.com/dotansimha/envelop/pull/498 , closing.

PR #354, opened 2021-07-01
improve disable introspection with conditional

PabloSzx:
Currently the disable introspection plugin didn't allow any customization / conditionals, for example, to only disable introspection if "X" authorization header is not present/allowed

I had to use `Parameters<NonNullable<Plugin['onValidate']>>[0]['params']` because the params type was inlined, maybe it could be worth it to refactor all the plugins options types into their own interfaces

### DIFF
--- a/.changeset/green-mails-tickle.md
+++ b/.changeset/green-mails-tickle.md
@@ -1,0 +1,5 @@
+---
+'@envelop/disable-introspection': minor
+---
+
+add optional conditional "disableIf" with context and params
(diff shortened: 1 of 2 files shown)


n1ru4l (approved):
looking good!

dotansimha (approved):

dotansimha:
> I had to use `Parameters<NonNullable<Plugin['onValidate']>>[0]['params']` because the params type was inlined, maybe it could be worth it to refactor all the plugins options types into their own interfaces

Yeah, planned for next week :)

PR #367, opened 2021-07-06
Experiment with better `context` type for plugins, with inferred context value 

dotansimha:
TODO: 
- [x] `infer` types where possible
- [x] Refactor `Plugin` type 
- [x] Refactor `GetEnvelopedFn`
- [x] Merge all context types together
- [x] Export a helper for `ContextOf<Envelop>`
- [x] Refactor hooks types 
- [x] Support module augmentation

Also, in this PR:
- [x] Improve `prepareSchema` and call it only when the schema changes, and not every time 
- [x] Cleanup for some traces of the tracing orchestrator implementation. 
  

The goal of this PR is to provide an amazing TypeScript experience for developers who are using TypeScript, by using type inference in order to understand what context extensions are coming from each plugin, what is the initial context passed to `getEnveloped`, and what's the context extension passes to `contextFactory`, and then build a unified TypeScript type for all of these:

image

image

Also, some plugins allow us to provide booleans in order to choose if to extend the context or not. We can infer the context type of the plugin by using `T extends true ? { something: true } : {}`. For example:

image

image


PabloSzx:
@dotansimha could be possible to also change 
```ts
export type DefaultContext = Record<string, unknown>;
```
 to 
```ts
export interface DefaultContext extends Record<string, unknown> {};
```

this enables module augmentation

dotansimha:
> ```ts
> export interface DefaultContext extends Record<string, unknown> {};
> ```

I'm also refactoring the hooks types, I want to achieve better types and more simple types. I hope to remove `DefaultContext` and replace it with something more robust. I will definitely export it as `interface` in order to allow module augmentation:)

PR #400, opened 2021-07-11
Extend Envelop plugin API with the ability to access and replace resolver function

dotansimha:
Related: https://github.com/dotansimha/envelop/issues/383


santino:
This is brilliant, thank you so much!

PR #410, opened 2021-07-12
Introduce useOverrideFields plugin

santino:
## Description

This PR implements a new plugin to override fields’ resolvers.

The goal is to replace the default resolver with something designed as a drop-in function that prevents the original resolver logic from being executed and eventually returns a simple value instead.
It is mostly meant to be for empty values, such as empty strings, null or undefined; but it can be fully customised to be anything more complex, such as returning values from a cache.

Fields can be targeted by plain string or Regular Expressions.

Targeted fields are always overridden by default, but custom logic can be implemented with the built-in `shouldOverride` function, to enable/disable resolver function replacement.

As usual, feedback welcome. Once we’re happy with the implementation I will complete documentation and add tests.

## Type of change

## Checklist:


packages/plugins/override-fields/README.md
```diff
@@ -0,0 +1,50 @@
...
+You can access subfields with a "dot" notation, as you would do when accessing properties of an object.
+
+```ts
+import { envelop } from '@envelop/core';
+import { useOverrideFields } from '@envelop/override-fields';
+
+const getEnveloped = envelop({
+  plugins: [
+    useOverrideFields({
+      fields: ['continents', 'country.callingCodes.countries', /country\.aplhaCode.*/],
```
dotansimha: I think this might need to be `Type.Field` instead of just field name? Otherwise you might effect types that you don't want to?

santino: Yes, you're right.
It is something I did think about during development, but then I forgot to implement.

Will add that

santino:
This PR is complete now, with documentation and tests.
Feel free to do a final round :)

dotansimha:
Available as alpha: `@envelop/override-fields@0.0.2-alpha-0f6ae12.0`

n1ru4l (commented):
To me, it seems like this is might actually better be considered as schema tooling and I am not sure whether it should be an envelop plugin or standalone schema transform/mapper implemented via https://www.graphql-tools.com/docs/api/modules/utils#mapschema The performance of using mapSchema will probably even be better 🤔

santino:
MapSchema is an interesting thought.
Potentially you can implement something to replace a resolver function with MapSchema and it would be done only once and commit the change to the schema.

If you want to use MapSchema directly, you'd need to build your matching logic and it would require more work and so become less intuitive.
The idea for a plugin is to be a solution that's ready to be dropped into your application.
As such, potentially this plugin could be "re-implemented" making use of MapSchema, instead of using Hooks from Envelop.
That is at the cost of introducing `@graphql-tools/utils` as a dependency in your application.

The drawback to using MapSchema is that the logic to determine when you want to enable/disable the override must run on each field that is matched.
This is a bit of a shame because the context and variables you send with your operation are unique to the operation only, and not to each field, and these are the values you're going to use to build your logic.
In my PR, in fact, [`shouldOverride`](https://github.com/dotansimha/envelop/pull/410/files#diff-13e49f8d9b79579f7a39bc0806babf651baef8419c8a096be6e94b85263874adR33) only runs once within the `onExecute` hook.
Then if the logic establishes that the override should not happen (returning `false`), nothing else happens; the. `onResolverCalled` won't be invoked.
Otherwise we're relying on `onResolverCalled` which is something used by many other plugins, and hopefully on its own doesn't introduce much overhead.
The implementation for matching the field name that lives within `onResolverCalled` is trivial, because it just compares strings and so won't impact runtime performance really.

santino:
Hello folks,
a gentle nudge on this one.

Would love to get some movement in here so that I know which direction we want to go with this plugin.

PR #424, opened 2021-07-15
refactor testing kit

n1ru4l:


packages/testing/src/index.ts
```diff
@@ -79,39 +80,93 @@ export function createTestkit(
...
-        ...args,
-        contextValue: await proxy.contextFactory(),
+      let document: DocumentNode;
+      try {
+        document = isDocumentNode(operation) ? operation : proxy.parse(operation);
+      } catch (err) {
+        return {
+          errors: [err],
+        };
+      }
```
n1ru4l, line 93: helix has custom logic for converting normal errors to graphql errors.

packages/testing/src/index.ts
```diff
@@ -79,39 +80,93 @@ export function createTestkit(
...
+            },
+          },
+          document,
+          operation: print(document),
+          variables: variableValues,
+          ...initialContext,
+        });
+      } catch (err) {
+        return {
+          errors: [err],
```
n1ru4l: I think we should do instanceof GraphQLError checks here and throw the original error if it does not extends a GraphQLError instead of returning it, @dotansimha what do you think? Helix seems to convert any error into a GraphQLError

dotansimha (approved):

dotansimha:
I guess `graphql-helix` can be removed as well from `package.json`?

packages/testing/src/index.ts
dotansimha: Yes! this should make it more simple to catch unexpected errors during tests

PR #432, opened 2021-07-16
feat: new execute-subscription-event plugin

n1ru4l:
Successor of https://github.com/dotansimha/envelop/pull/183
Closes https://github.com/dotansimha/envelop/issues/80

### DIFF
--- a/packages/plugins/execute-subscription-event/src/subscribe.ts
+++ b/packages/plugins/execute-subscription-event/src/subscribe.ts
@@ -1,0 +1,49 @@
+import { createSourceEventStream } from 'graphql';
+
+import { ExecuteFunction, makeSubscribe, SubscribeFunction } from '@envelop/core';
+import isAsyncIterable from 'graphql/jsutils/isAsyncIterable';
+import mapAsyncIterator from 'graphql/subscription/mapAsyncIterator';
+
+/**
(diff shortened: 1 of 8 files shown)


packages/plugins/execute-subscription-event/src/subscribe.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { ExecuteFunction, makeSubscribe, SubscribeFunction } from '@envelop/core';
+import isAsyncIterable from 'graphql/jsutils/isAsyncIterable';
+import mapAsyncIterator from 'graphql/subscription/mapAsyncIterator';
+
+/**
+ * This is a almost identical port from graphql-js subscribe.
+ * The only difference is that a custom `execute` function can be injected for customizing the behavior.
+ */
+export const subscribe = (execute: ExecuteFunction): SubscribeFunction =>
```
n1ru4l, line 11: > @dotansimha [original comment on old pr]( https://github.com/dotansimha/envelop/pull/183#discussion_r661352460)
Some plugins might use setExecuteFn during onExecute with the assumption that they override this function.
Does it mean that these plugins need somehow to override this function here?

I don't think so, at least for now. I cannot think of a use-case where you would want a custom `execute` function for `subscribe` right now 🤔

dotansimha (approved):

PR #462, opened 2021-07-22
fix typings

n1ru4l:


PabloSzx:
those "as any" in the tests are something to be required to use "setExecuteFn" and "setSubscribeFn"?

n1ru4l:
@PabloSzx yes, I don't understand the typescript error. Maybe you can decrypt it for me? 😅

PabloSzx:
The issue seems to be related with TypeScript itself, created an issue in the TypeScript Repo that should be related to this https://github.com/microsoft/TypeScript/issues/45153


image

PabloSzx (approved):

PabloSzx:
image

Changing the execute & subscribe type definitions of the graphql.js package to use tuples instead of overloading a function fixes all the issues, ~~I feel like this fix should be made there~~ The change for graphql is already in v16, where it's only possible to give execute & subscribe an object with the arguments

n1ru4l:
Seems like we need to release a v2 in about one year or so when graphql 16 is ready 😄 

@PabloSzx Thanks for the research ❤️ 

@dotansimha Feel free to merge!

PR #474, opened 2021-07-23
resource-limitations plugin

n1ru4l:
Related https://github.com/dotansimha/envelop/issues/464

Apply extended validation rules, based on https://docs.github.com/en/graphql/overview/resource-limitations


dotansimha:
I love it. Just missing a changeset and we can continue:)
Do you think we should allow to customize the suffixes/ input field names?

dotansimha:
~~Also, should it maybe be in a separate package?~~

n1ru4l:
@dotansimha outside the mono repo?

dotansimha:
Oh nevermind, I thought it was part of extended-validation plugin :)

dotansimha:
And maybe we can add another plugin for rate limit based on points / complexity ? (see https://shopify.engineering/rate-limiting-graphql-apis-calculating-query-complexity)

n1ru4l:
@dotansimha Yes lets go 🔥🔥🔥🚀

I also think we should maybe move oneOf to a standalone plugin

PR #491, opened 2021-07-26
Added `handleStreamOrSingleExecutionResult ` helper

dotansimha:
In this PR:
- Added `handleStreamOrSingleExecutionResult` helper for making it simpler to deal with stream responses.
- Added documentation on how to implement `onExecuteDone` hook.
- Moved `isAsyncIterable` into our repo, to make it work with all versions of GraphQL.
- Refactor plugins to use `handleStreamOrSingleExecutionResult`.

### DIFF
--- a/packages/plugins/newrelic/src/index.ts
+++ b/packages/plugins/newrelic/src/index.ts
@@ -1,8 +1,7 @@
 import { shim as instrumentationApi } from 'newrelic';
-import { Plugin, OnResolverCalledHook } from '@envelop/types';
+import { Plugin, OnResolverCalledHook, isAsyncIterable } from '@envelop/types';
 import { print, FieldNode, Kind, OperationDefinitionNode } from 'graphql';
 import { Path } from 'graphql/jsutils/Path';
-import isAsyncIterable from 'graphql/jsutils/isAsyncIterable.js';
 
 enum AttributeName {
   COMPONENT_NAME = 'Envelop_NewRelic_Plugin',
--- a/packages/types/src/async-utils.ts
+++ b/packages/types/src/async-utils.ts
@@ -1,7 +1,25 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 import { OnExecuteDoneEventPayload, OnExecuteDoneHookResult, OnExecuteDoneHookResultOnNextHook } from '@envelop/core';
-import isAsyncIterable from 'graphql/jsutils/isAsyncIterable.js';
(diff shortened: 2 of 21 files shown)


packages/types/src/async-utils.ts
```diff
@@ -0,0 +1,18 @@
...
+export function handleMaybeStream(
+  payload: OnExecuteDoneEventPayload,
+  fn: OnExecuteDoneHookResultOnNextHook
+): void | OnExecuteDoneHookResult {
+  if (isAsyncIterable(payload.result)) {
+    return { onNext: fn };
+  } else {
+    fn({
+      result: payload.result,
+      setResult: payload.setResult,
```
dthyresson, line 31: Is `setResult` optional?

In the case with our custom Logger plugin in RedwoodJS, we want to log portions of the `result` -- options redacted data and some timing traces -- in `onExecuteDone` but not necessarily modify the result.

n1ru4l: `setResult` is injected into the hook but fully optional. You don't need to call it, only if you want to override the execution result.

packages/types/src/async-utils.ts
```diff
@@ -0,0 +1,18 @@
+import { OnExecuteDoneEventPayload, OnExecuteDoneHookResult, OnExecuteDoneHookResultOnNextHook } from '@envelop/core';
+import isAsyncIterable from 'graphql/jsutils/isAsyncIterable.js';
+
+export function handleMaybeStream(
```
n1ru4l: `handleStreamOrSingleExecutionResult` ?

Also, we can add a jsdoc on why it is useful - with a short usage example?

packages/plugins/newrelic/src/index.ts
```diff
@@ -1,8 +1,7 @@
 import { shim as instrumentationApi } from 'newrelic';
-import { Plugin, OnResolverCalledHook } from '@envelop/types';
+import { Plugin, OnResolverCalledHook, isAsyncIterable } from '@envelop/types';
```
n1ru4l, line 2: exporting this from `@envelop/types` makes it no longer a type-only package 🤔  Should we move it to the core instead?

dotansimha: I wanted to put that in the core, but then it means that the plugins know the core, which might not be ideal? I know the name `types` is confusing in that context, but it's more like a `common` is `shared` now 

I tried to keep it as:

```
core (direct) -> types
plugins (indirect) -> types
```

PR #498, opened 2021-07-27
Refactoring of Persisted Operations plugin

santino:
## Description

This PR takes input from #324 to refactor the existing Persisted Operations plugin to bring improvements initially proposed in the other PR as a new plugin.

The end goal is to implement the changes to the existing plugin rather than raising a new one.
As such #324 has served as a conversation starter and will be eventually closed in favor of this.

## Type of change

## Changes introduced to the current plugin:
- Write `operationId` in context as plain property (no symbol) so it can be reused by other plugins (e.g. logging)
- Prescribe store shape, Map of key (operationId) value (stringified query or DocumentNode)
- Removed `canHandle` and `get`, since we can implement their logic for free now that we know store shape
- Setting the operation id from a custom property within context
- Support multiple lists by passing a store function that returns a single store to use per operation:`` store: context => persistedOperationsStore.get(`${context.request.header.clientName}_persistedOperations`) ``
- Live updates to the store; a store can be implemented through a class that has methods to update the store during runtime, see [jsonFilesStore](https://github.com/dotansimha/envelop/pull/498/files#diff-91c23a63c544e0d29b22e6a098ae74fc403ec9d6aac5babe3806fc30d1ade659) as an example. The plugin will retrieve the up to date instance of the store Map during operation Parse
- Added sample store implementation to handle multiple static files, [jsonFilesStore](https://github.com/dotansimha/envelop/pull/498/files#diff-91c23a63c544e0d29b22e6a098ae74fc403ec9d6aac5babe3806fc30d1ade659). This serves as an example to manage multiple lists and live updates to the stores. Supports single json file without passing the listName, or multiple files requiring to pass which of the available lists (files) to read from to fulfill the operation

## Additional proposals, not yet implemented as they would be breaking:
- Remove `writeToContext` boolean and always write to context, cannot see any damage in doing this
- Rename `onlyPersistedOperations` to `onlyPersisted` to be less verbose

## Checklist:


packages/plugins/persisted-operations/src/jsonFilesStore.ts
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable no-console */
+import { basename, extname, isAbsolute, resolve } from 'path';
+import { readFile } from 'fs/promises';
+import { PersistedOperationsStore } from './plugin';
+
+export class JsonFilesStore {
```
n1ru4l: To me, this class feels way more complicated and confusing than it must be. A Store on which you call a get method to get a Store?

I don't think we need to provide a primitive for loading multiple JSON files and should be up to the user to implement.

Something basic like the following could be more straightforward.

```ts
class JSONStore extends Map {
  async load(path: string) {
    // logic for loading path
   const data = await loadJSONFile(path)
   for (const [key, value] of Object.entries(operationsObject)) {
      this.set(key, value);
    }
  }
}
```

With the following usage:

```ts
const persistedOperationsPaths = [
  resolve(process.cwd(), 'assets/client1PersistedOperations.json'),
  resolve(process.cwd(), 'assets/client2PersistedOperations.json'),
];

const store = new JSONStore();
for (const path of persistedOperationsPaths) {
  store.load(path);
}
 // load and parse persisted-operations files

  plugins: [
    // ... other plugins ...
    usePersistedOperations({
      store,
    }),
  ],
});
```

And if you need to use a different store, based on context:

```ts
const persistedOperationsPaths = [
  resolve(process.cwd(), 'assets/client1PersistedOperations.json'),
  resolve(process.cwd(), 'assets/client2PersistedOperations.json'),
];

const store1 = new JSONStore();
const store2 = new JSONStore();

store1.load(path[0]);
store2.load(path[1]);
 // load and parse persisted-operations files

  plugins: [
    // ... other plugins ...
    usePersistedOperations({
      store: ctx => ctx.sth ? store1 : store2,
    }),
  ],
});
```

santino: > this class feels way more complicated and confusing than it must be

Possible, but my goal with this class is not just to be something that people can use, but also to give examples of how you can build full solutions with more complex scenarios, such as handling multiple lists with a clue on how you can update them (all or some) on the fly.
Even though, yeah, maybe you can achieve that with a much simpler class, at the cost of actually having to initiate multiple instances to serve multiple stores.

Maybe we can commit a more minimal setup and this more complex solution could live in a guide or a recipe.

> A Store on which you call a get method to get a Store?

To be more precise, a store which you call a get method to get the Map of persisted operations that are relevant to the requested operation, which might be only from a single list of those supported by your store.

> should be up to the user to implement.

Yes, but don’t forget that you don’t have to use this class in order to use the plugin.
Users are still free to implement what works for them.
This works, it handles a single list scenario, more complex scenarios with multiple lists and live updates; but it's also there as a reference.

> Something basic like the following could be more straightforward

It would.
But personally, I would not extend from Map.
At that stage you return a Map that people can mutate directly by invoking all the methods available, including `set`; the risk is much bigger as you might end up with something that goes well beyond storing entries from a designated static file.
This is also the reason why in my implementation I have a method `get` that returns the Map, and I keep the store as private; I want to enforce immutability patterns and make it straightforward that you retrieve the Map and so this is meant to be read-only.

Anyway, I am not precious about the jsonFilesStore.
If the feedback is to have something that is much simpler and we fully leave it to the users to implement more complex scenarios on their own, I will be ok with that.
The more complex store implementation might well live in a guide or blog post, just to inspire or provide some guidance.

dotansimha (commented):
I pushed some changes, mainly around:
- The store shouldn't be a `Map`, it should be an interface with `.get` method. This is needed for flexibility. 
- The stores shouldn't handle multiple stores. I added `AggregatedStore` in cases where you might want to group stores together for some reason.
- There is the ability to choose the store in a flexible way. This should address the conditional stores/choosing stores dynamically.

@santino what do you think?

santino:
Thanks for your review and the changes.
It looks good and I like the idea of throwing in even more stores implementations.

I wanted to ask why you went back with a Symbol for the context key.
Realistically if people want to write in context (as they pass the flag explicitly), is because they want to access the value from context.
In this case should make it simple to do so?

Also, what do you think of the additional proposals as per description?
> - Remove writeToContext boolean and always write to context, cannot see any damage in doing this
> - Rename onlyPersistedOperations to onlyPersisted to be less verbose

dotansimha:
> I wanted to ask why you went back with a Symbol for the context key.
> Realistically if people want to write in context (as they pass the flag explicitly), is because they want to access the value from context.

To make sure it's not accessible at all without the helper function, and to make sure it's never going to conflict with other fields that are in the `context` (`operationId` is pretty common...). 
If someone needs it, it's possible to get it using the helper function. 


> * Remove writeToContext boolean and always write to context, cannot see any damage in doing this

Agree!

> > * Rename onlyPersistedOperations to onlyPersisted to be less verbose

Agree!

santino:
> If someone needs it, it's possible to get it using the helper function.

Fair!

Your changes look good to me.
Looking forward for this new version of the plugin to be released!

dotansimha:
Thanks @santino ❤️ . We are good to go. Note that I changed `setOperationId` to be `extractOperationId`, and updated the README.

PR #499, opened 2021-07-27
Correct usage example

dburles:
This is a pretty minor change to the readme. This was difficult to debug as I am not using Typescript and there are no runtime checks of what's passed into the `useDataLoader` function.


dotansimha:
Thank you!

PR #618, opened 2021-08-26
fix(useMaskedErrors): handle subscription errors

n1ru4l:


n1ru4l:
Added changes over in https://github.com/dotansimha/envelop/pull/617/files

PR #675, opened 2021-09-13
docs: new guide Adding Authentication with Auth0

n1ru4l:


examples/graphql-helix-auth0/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      String that uniquely identifies an authenticated user.
+      """
+      sub: String!
+    }
+
+    type Query {
+      """
+      The authentication information of the request.
+      """
+      authInfo: AuthenticationInfo
```
dotansimha, line 24: Maybe this could be `type Query { viewer: User! } type User { id: ID! }` (and then map it to `auth0.sub`) ?

examples/graphql-helix-auth0/index.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+const getEnveloped = envelop({
+  plugins: [
+    useSchema(schema),
+    useAuth0({
+      domain: auth0Config.domain,
+      audience: auth0Config.audience,
+      extendContextField: 'auth0',
+    }),
```
dotansimha, line 49: Maybe we can add another example for adding `useGenericAuth` as well? (it can be combined with the `context.auth0` to resolve/validate the user) and also show how you can add auth using `@auth` directive?

dotansimha: but it's a separate example, not sure if it's too advanced

packages/plugins/auth0/README.md
```diff
@@ -6,11 +6,9 @@ This plugin validates an JWT token created by [Auth0](https://auth0.com/), and i
 
 ## Getting Started
 
-```
-yarn add @envelop/auth0
-```
+We recommend using the [Adding Authentication with Auth0 guide](/docs/guides/adding-authentication-with-auth0) if this is your first time using this plugin!
```
dotansimha: I think this link might not work in GitHub README?

website/docs/guides/adding-authentication-with-auth0.mdx
```diff
@@ -0,0 +1,286 @@
+---
+title: Adding Authentication with Auth0
+sidebar_label: Adding Authentication with Auth0
+---
```
dotansimha, line 4: Maybe add a note that Auth0 has many ways of integrating their client SDK, and in this example, you are using it from a CDN for simplicity. 
And you can link to different integrations (for example https://github.com/auth0/nextjs-auth0 )

dotansimha (approved):

examples/graphql-helix-auth0/index.ts
n1ru4l: I think that should be a separate example.

website/docs/guides/adding-authentication-with-auth0.mdx
n1ru4l: https://github.com/n1ru4l/envelop/blob/ad6450ee2aa9de5f4e3e74ee5e156f000a822070/website/docs/guides/adding-authentication-with-auth0.mdx#L170-L171

examples/graphql-helix-auth0/index.ts
n1ru4l: I wanted to reduce complexity by avoiding doing complex mappings within the resolvers. At the end in the `Next steps` section, I briefly mention this.

PR #685, opened 2021-09-14
@envelop/response-cache Implements Redis Cache

### LINKED ISSUE #669: @envelop/response-cache: Add a Redis backed cache
Currently, the only type of cache available to the `@envelop/response-cache` plugin is an In-Memory Cache.

Such a cache isn't practical for RedwoodJS applications when using a serverless (AWS Lambda) function for its the GraphQL handler.

Instead, if a Redis store as a "global" cache provided by a service like Upstash [https://www.upstash.com](https://www.upstash.com) as available, the GraphQL execution results could be stored as keys for each responseId and could also be deleted on invalidation.

* Instead of implementing and distribution the Redis-cache, document how to implement it and developers could create their own following the Cache interface

**Additional context**

The docs/readme:

* Implement your own global cache (e.g. using Redis) by implementing the Cache interface.

dthyresson:
## Description

Fixes #669

Implements a Redis cache for use with the @envelop/response-cache plugin.

This cache implements the set, get, and invalidate methods necessary for storing the Execution result response an associating the response with the entities it contains (and vice versa).

This cache uses the `ioredis` client: see: https://github.com/luin/ioredis

#### Get / Query Entity

A query to fetch Post with id 1 would

* cache the result of the query 

```
data: { post: id: 1. title: "My Post" } }
``` 

and associate with a generated response id (for example (`abcd12345xyz=`)
* associate `Post:1` with that response id `abcd12345xyz=`
* associate the response with a set of entities it contains `abcd12345xyz=` with `Post:1`
* But since `abcd12345xyz=` is a key with a value of the response, those associations are in a key `operations:abcd12345xyz=` with `Post:1` and `Post` as a members

#### Get / QueryAll Entities

A query to fetch all Comments

* cache the result of the query 
 ```
data: 
  { 
       { comment: id: 1. title: "Great post!",  posts: { id: 1 } }, 
       { comment: id: 2. title: "I agree 100%", posts: { id: 1 } }
  }
``` 
and associate with a generated response id (for example (`fghi6789abc=`)
* associate `Comment:1` and `Comment:2`  with that response id `fghi6789abc=`
* associate the response with a set of entities it contains `fghi6789abc =` with `Comment:1` and `Comment:2`
* But since `fghi6789abc =` is a key with a value of the response, those associations are in a key `operations:fghi6789abc =` with `Comment:1`, `Comment:2` and `Comment` as a members

#### Mutations
Mutations will invalidate and remove the responses associated with the entity ... and the entity from the any other responses and those responses.

#### Invalidations

Invalidations can also be called in database middleware (like Prisma) by using the same cache that would be set on the plugin

```ts
 import Redis from 'ioredis';

 const redis = new Redis();
 const cache = createRedisCache({ redis });

// ...

useResponseCache({ cache })

// ...
// @See: Examples of how to generate identifiers to invalidate.
cache.invalidate(entity)
```
#### Creating Invalidation Entities

```ts

const redis = new Redis() // with connection string or RedisOptions;
const cache = createRedisCache({ redis });

const identifier = new Map<string, CacheEntityRecord>()

const cachedEntityRecord = <CacheEntityRecord>{
  typename: 'Post`,
  id: 1,
}

 identifier.set('Post' cachedEntityRecord)

cache.invalidate(identifier.values())
```

If doing invalidation via Prisma middleware:

```ts
db.$use(async (params, next) => {
  const result = await next(params)

  if (
    params.action == 'create' ||
    params.action == 'createMany' ||
    params.action == 'delete' ||
    params.action == 'deleteMany' ||
    params.action == 'update' ||
    params.action == 'updateMany' ||
    params.action == 'upsert'
  ) {
    logger.debug(
      {
        model: params.model,
        action: params.action,
      },
      'Primsa middleware'
    )

    const identifier = new Map<string, CacheEntityRecord>()
    const cachedEntityRecord = <CacheEntityRecord>{
      typename: params.model.toString(),
      id: result.id,
    }
    logger.debug(
      { cachedEntityRecord: cachedEntityRecord },
      'cached record to delete'
    )

    identifier.set(params.model.toString(), cachedEntityRecord)

    logger.debug({ values: identifier.values() }, 'keys to invalidate')

    cache.invalidate(identifier.values())
  }

  return result 
})

```

### Creating client and cache

One would create the Redis client using a connection string like:

`redis://something@us1-some-instance.upstash.io:port`

Or via a set of `RedisOptions., see: https://github.com/luin/ioredis/blob/master/lib/redis/RedisOptions.ts when [Connecting to Redis](https://github.com/luin/ioredis#connect-to-redis)

```ts
const redis = new Redis() // with connection string or RedisOptions;
const cache = createRedisCache({ redis });
```

Note that:

* associations are Redis `sets` https://redis.io/commands#set where the key has a set of members 
* responses are key values where the value is the execution result store as JSON

## Further comments

Uncertain of:

* Invalidation logic
* If invalidation is async such that when invalidating, the response is fetched before invalidation is complete thus the response contains old data
* If the "base" typename for example "Post" should invalidate all

Other

* a helper/utility to create the invalidate identifiers could be useful so just pass a typename and id or an array of then that creates `cache.invalidate(identifier.values())`

## Type of change

Adding links to sandbox or providing screenshots can help us understand more about this PR and take action on it as appropriate

## How Has This Been Tested?

The `response-redis-cache.spec` test suite 

- [x] response-redis-cache.spec

**Test Environment**:

- OS: OSX
- `@envelop/response-cache 0.2.0`:
- NodeJS:

## Checklist:

- [X] I have followed the [CONTRIBUTING](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md) doc and the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented on my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


kamilkisiela:
Maybe it's worth to use pipelining if possible or reduce `store.del` to one request.

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+  buildRedisResponseOpsKey?: BuildRedisResponseOpsKey;
+};
+
+export const createRedisCache = (params: RedisCacheParameter): Cache => {
+  const store = params.redis;
+
+  const buildRedisEntityId = params?.buildRedisEntityId ?? defaultBuildRedisEntityId;
+  const buildRedisResponseOpsKey = params?.buildRedisResponseOpsKey ?? defaultBuildRedisResponseOpsKey;
+
+  function purgeEntity(entity: string) {
```
kamilkisiela: Maybe this could return an array of keys to delete?

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+    // find the responseIds for the entity
+    store.smembers(entity).then(function (responseIds) {
+      // and purge each response since they contained the entity data
+      store.del(
+        responseIds.map(responseId => {
+          return buildRedisResponseOpsKey(responseId);
+        })
+      );
+      store.del(responseIds);
```
kamilkisiela: Instead of deleting records, we could push to an array of keys.

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+                return buildRedisResponseOpsKey(responseId);
+              })
+            );
+            // and purge each response since they contained the entity data
+            store.del(responseIds);
+          });
+        }
+
+        // then purge the entityKeys like Comment:1, Comment:2 etc
+        store.del(entityKeys);
```
kamilkisiela: Here we could push to an array of keys as well.

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+      });
+    }
+
+    // and then purge the entity itself
+    store.del(entity);
```
kamilkisiela: again, push and 

```typescript
return keys;
```

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+        store.sadd(responseKey, typename);
+
+        if (id) {
+          const entityId = buildRedisEntityId(typename, id);
+          // Adds a key for the typename:id => response
+          store.sadd(entityId, responseId);
+          // Adds a key for the operation => typename:id
+          store.sadd(responseKey, entityId);
+        }
+      }
```
kamilkisiela:
```suggestion
      const pipeline = store.pipeline();
      if (ttl === Infinity) {
        pipeline.set(responseId, JSON.stringify(result));
      } else {
        // set the ttl in milliseconds
        pipeline.set(responseId, JSON.stringify(result), 'PX', ttl);
      }

      const responseKey = buildRedisResponseOpsKey(responseId);

      for (const { typename, id } of collectedEntities) {
        // Adds a key for the typename => response
        pipeline.sadd(typename, responseId);
        // Adds a key for the operation => typename
        pipeline.sadd(responseKey, typename);

        if (id) {
          const entityId = buildRedisEntityId(typename, id);
          // Adds a key for the typename:id => response
          pipeline.sadd(entityId, responseId);
          // Adds a key for the operation => typename:id
          pipeline.sadd(responseKey, entityId);
        }
      }
      
      pipeline.exec()
```

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+      return (
+        store.get(responseId).then(function (result) {
+          return result && JSON.parse(result);
+        }) ?? null
+      );
+    },
+    invalidate(entitiesToRemove) {
+      for (const { typename, id } of entitiesToRemove) {
+        purgeEntity(id != null ? buildRedisEntityId(typename, id) : typename);
+      }
```
kamilkisiela:
```suggestion
      const keys$: Promise<string[]>[] = [];
      for (const { typename, id } of entitiesToRemove) {
        keysPromises.push(purgeEntity(id != null ? buildRedisEntityId(typename, id) : typename));
      }
      
      Promise.all(keys$).then(keys => store.del(keys.flat()));
```

kamilkisiela (commented):
I added few comments on how we could reduce the number of calls to redis.

packages/plugins/response-cache/src/redis-cache.ts
kamilkisiela: diff looks awkward... I used`store.pipeline()` replaced `store.sadd` with `pipeline.sadd` and added `pipeline.exec()` at the end.

packages/plugins/response-cache/src/redis-cache.ts
dthyresson: @kamilkisiela Great idea. I actually had that implemented in an earlier version and then reverted to individual commands -- but that optimization is better. I like returning a list of keys, too.

I'll make that update. Thanks!

packages/plugins/response-cache/src/redis-cache.ts
dthyresson: I also had thought of ... well a transaction to surround the adds ... but pipeline is better. Will implement.

packages/plugins/response-cache/src/redis-cache.ts
dthyresson: Resolves promises. Though instead of ` Promise<string[]>[]`  the purgeEntity returns  Promise<string[]>

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,127 @@
...
+  buildRedisResponseOpsKey?: BuildRedisResponseOpsKey;
+};
+
+export const createRedisCache = (params: RedisCacheParameter): Cache => {
+  const store = params.redis;
+
+  const buildRedisEntityId = params?.buildRedisEntityId ?? defaultBuildRedisEntityId;
+  const buildRedisResponseOpsKey = params?.buildRedisResponseOpsKey ?? defaultBuildRedisResponseOpsKey;
+
+  function purgeEntity(entity: string): Promise<string[]> {
```
dthyresson: I could see renaming `purgeEntity` to something more like `buildEntityInvalidationsKeys` since it does not actually "purge" or execute the `del` commands.

Is this worth renaming?

kamilkisiela: Makes sense 👍🏻

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,127 @@
...
+    return new Promise(function (resolve, reject) {
+      const keysToPurge: string[] = [entity];
+
+      // find the responseIds for the entity
+      store.smembers(entity).then(function (responseIds) {
+        // and purge each response since they contained the entity data
+        responseIds.forEach(responseId => {
```
dthyresson: I chose to push each key, though one could push the array of keys and then flatten at the end ... but since I have to iterate in order to construct the operations key, figured was simpler to push each.

packages/plugins/response-cache/src/redis-cache.ts
```diff
@@ -0,0 +1,127 @@
...
+
+            // then purge the entityKeys like Comment:1, Comment:2 etc
+            entityKeys.forEach(entityKey => {
+              keysToPurge.push(entityKey);
+            });
+          });
+        }
+
+        // and then purge the entity itself
+        resolve(keysToPurge);
```
kamilkisiela: This might be buggy. The `resolve(keysToPurge)` does not wait for all the async actions above.

dthyresson: Oh. That's not good -- or what I want. Could you recommend another approach?

dthyresson: Changed logic in `buildEntityInvalidationsKeys` such that *I think* all paths will resolve.

dthyresson: Actually, I don't think my approach is correct. Revisiting.

dthyresson: Have changed to use async/await which I prefer, but can change if otherwise.

dthyresson:
@kamilkisiela  Might we want some helper to construct the invalidation identifiers?

For example, when I tried using Prisma middleware as:

```ts
db.$use(async (params, next) => {
  const result = await next(params)

  if (
    params.action == 'create' ||
    params.action == 'createMany' ||
    params.action == 'delete' ||
    params.action == 'deleteMany' ||
    params.action == 'update' ||
    params.action == 'updateMany' ||
    params.action == 'upsert'
  ) {
    logger.debug(
      {
        model: params.model,
        action: params.action,
      },
      'Primsa middleware'
    )

    const identifier = new Map<string, CacheEntityRecord>()
    const cachedEntityRecord = <CacheEntityRecord>{
      typename: params.model.toString(),
      id: result.id,
    }
    logger.debug(
      { cachedEntityRecord: cachedEntityRecord },
      'cached record to delete'
    )

    identifier.set(params.model.toString(), cachedEntityRecord)

    logger.debug({ values: identifier.values() }, 'keys to invalidate')

    cache.invalidate(identifier.values())
  }

  return result 
})
```

You can see that I have to assemble a map of `CacheEntityRecord`s and then pass those `.values()` to the cache's `invalidate`:

```
const entities = new Map<string, CacheEntityRecord>()

const cachedEntityRecord = <CacheEntityRecord>{
  typename: params.model.toString(),
  id: result.id,
}

entities(params.model.toString(), cachedEntityRecord)

cache.invalidate(entities.values())
```

Though not 100% sure how to simplify. 

Maybe I just document manual invalidation instead?

Although, I just noticed that the `Invalidate Cache based on custom logic` section has which is much more straightforward:

```
  cache.invalidate([
    {
      typename: resource.type,
      id: resource.id,
    },
  ]);
```

dthyresson:
@kamilkisiela I just learned about creating change sets. You this PR be a minor or major bump? I expect just a minor one.

n1ru4l:
@dthyresson I think the probably best way of shipping the redis cache implementation would be to create a new package within the monorepo with the name `@envelop/response-cache-redis`. That way you dont end up with redis dependency and code if you do not want to use redis 🤔 . Opinions, @dotansimha and @kamilkisiela ?

kamilkisiela (approved):
Creating a separate package for redis would be nice, I guess we may want to use other stores in future.

dthyresson:
> I think the probably best way of shipping the redis cache implementation would be to create a new package within the monorepo with the name @envelop/response-cache-redis.

> Creating a separate package for redis would be nice, I guess we may want to use other stores in future.

I can definitely refactor to have `@envelop/response-cache-redis`.

Would this package depend on `@envelop/response-cache`? Is it a separate plugin? And then all "core" response-cache functionality is effectively imported from `@envelop/response-cache`?

Or, is `@envelop/response-cache-redis` just the Redis cache itself?

Meaning, would one import:

* @envelop/response-cache (and)
* @envelop/response-cache-redis

Then use `response-cache-redis` just to

```ts
const redis = new Redis(("rediss://:1234567890@my-redis-db.example.com':30652");
 const cache = createRedisCache({ redis });
   plugins: [
     // ... other plugins ...
     useResponseCache({ cache }),
   ],
 });
```

Where `useResponseCache` is the plugin and `createRedisCache` comes from `response-cache-redis`?

dthyresson:
@kamilkisiela and @n1ru4l I refactored to make the cache itself a package and think I left the response-cache in its original state -- I removed the changeset and adjusted the docs a little. Though I am not sure if there is a version change needed or not.

As for the response-cache-redis package, I set it to version 0.1.0 and did not have a change set. I assume that is because it's new.

However, I am not sure how publishing the packages work so a review here is very much welcome.

As for the failing benchmark, I did run locally:

```
running (0m50.0s), 00/40 VUs, 271153 complete and 0 interrupted iterations
graphql-js                     ✓ [==================================] 10 VUs  10senvelop-just-cache             ✓ [==================================] 10 VUs  10s
prom-tracing                   ✓ [==================================] 10 VUs  10s
envelop-cache-and-no-intern... ✓ [==================================] 10 VUs  10s
envelop-cache-jit              ✓ [==================================] 10 VUs  10s
     ✓ no_errors
     ✓ expected_result
```

So may just a blip?

dthyresson:
Also, since looking at https://envelop-mcnwade84-theguild.vercel.app/plugins/use-response-cache and that is the plugin, perhaps I should have the docs make mention of the `@envelop/response-cache-redis` package and include the Redis examples there, too?

packages/plugins/response-cache-redis/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+export type RedisCacheParameter = {
+  /**
+   * Redis instance
+   * @see Redis.Redis https://github.com/luin/ioredis
+   */
+  redis: Redis.Redis;
+  /**
+   * Maximum amount of items in the cache. Defaults to `Infinity`.
+   */
+  max?: number;
```
n1ru4l: This parameter is unused, so we can remove it

packages/plugins/response-cache-redis/src/redis-cache.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Customize how the cache entity id is built.
+   * By default the typename is concatenated with the id e.g. `User:1`
+   */
+  buildRedisEntityId?: BuildRedisEntityId;
+  /**
+   * Customize how the cache key that stores the operations associated with the response is built.
+   * By default `operations` is concatenated with the responseId e.g. `operations:arZm3tCKgGmpu+a5slrpSH9vjSQ=`
+   */
+  buildRedisResponseOpsKey?: BuildRedisResponseOpsKey;
```
n1ru4l: This wording feels more clear to me `buildRedisOperationResultCacheKey`

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
```diff
@@ -0,0 +1,1155 @@
...
+    const comment2Members = await redis.smembers('Comment:2');
+    expect(comment2Members).toHaveLength(1);
+
+    await cache.invalidate([{ typename: 'Comment', id: 2 }]);
+
+    const commentMembersInvalidated = await redis.smembers('Comment');
+    expect(commentMembersInvalidated).toHaveLength(1);
+
+    const comment1MembersInvalidated = await redis.get('Comment:1');
+    expect(comment1Members).toHaveLength(1);
```
n1ru4l: It seems like this assertion should be `expect(comment1MembersInvalidated).toHaveLength(1);`?

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
```diff
@@ -0,0 +1,1155 @@
...
+        users {
+          id
+          name
+          comments {
+            id
+            text
+          }
+        }
+      }
+    `;
```
n1ru4l: This opertaion is unused

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
```diff
@@ -0,0 +1,1155 @@
...
+
+    const responseCache = result['extensions']['responseCache'];
+    const invalidatedEntities = responseCache['invalidatedEntities'];
+    expect(invalidatedEntities).toHaveLength(1);
+
+    const invalidatedUser = invalidatedEntities[0];
+    expect(invalidatedUser).toHaveProperty('typename');
+    expect(invalidatedUser).toHaveProperty('id');
+    expect(invalidatedUser.typename).toEqual('User');
+    expect(invalidatedUser.id).toEqual('1');
```
n1ru4l: instead of all those seperate assertions this would be a shorter one that covers all:

```
expect(invalidatedUser).toEqual({
  typename: "User",
  id: "1"
})
```

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
```diff
@@ -0,0 +1,1155 @@
...
+    `;
+
+    // query and cache
+    await testInstance.execute(query);
+    // from cache
+    await testInstance.execute(query);
+    // so queried just once
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    await new Promise(resolve => setTimeout(resolve, 150));
```
n1ru4l: Could this be done by advancing fake timers instead?

packages/plugins/response-cache-redis/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@envelop/response-cache-redis",
+  "version": "0.1.0",
```
n1ru4l: You can set this to `0.0.0` and create a changeset for the initial release. The changeset action will then take care of bumping the version.

packages/plugins/response-cache-redis/src/redis-cache.ts
dthyresson: Right, the max was helpful in the LRU in-memory-cache but not necessarily needed here. Imagine if too many keys (w/out expiration) happens in the reds store, one could flush if started to approach max storage.

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
dthyresson: Yes, you are correct. I actually I found a bug  (a get vs smembers) in that test case and have corrected as well as added code comments to make it a little more clear what is happening and being verified.

packages/plugins/response-cache-redis/test/response-redis-cache.spec.ts
dthyresson: I used the same test that existed in the LRU cache-backed spec. But, I'll look at this. I think I saw fake timers used in another package.

dthyresson: I think thought, that this test does want to really wait 150 msec so the item expires. But perhaps there is a better way to test that not using "real-time".

dthyresson:
@n1ru4l Thanks for the review and I believe I have made the necessary changes and guides/documentation updates suggested. I am not sure if the changesets are bumping appropriately but did create of for the -redis package (and the response-cache should just have readme changes?).

website/docs/guides/adding-a-graphql-response-cache.md
```diff
@@ -298,26 +298,45 @@ const getEnveloped = envelop({
 });
 ```
 
-Want a global cache on Redis? Build a cache that implements the `Cache` interface and share it with the community!
+Want a global cache on Redis?
+
+Maybe you are in a server-less environment and the In-Memory Cache isn't an option.
```
n1ru4l: Maybe we can also add a note about server replica deployments?

```suggestion
Maybe you are in a server-less environment and the In-Memory Cache isn't an option. Also, when having multiple server replicas, you might want to have a shared cache between all the replicas.
```

n1ru4l:
@dthyresson Just a small doc change to also mention replicas, then we can merge and release it :)

dthyresson:
Thanks @n1ru4l - added the docs suggested that mention replicas.

PR #687, opened 2021-09-15
feat: firebase-auth plugin

saihaj:
A simple plugin that allows to use [Firebase Auth](https://firebase.google.com/docs/auth) to authenticate users. 

**Flow for this plugin:**
Client logs in using client SDK -> Frontend will pass a JWT that is used to identify the user to a Firebase service -> this plugin extracts the headers for that JWT and verifies the identity -> if all good, firebase user object is added to context.

I made a demo here https://github.com/saihaj/firebase-graphql-envelop that only allows authenticated users. 

**To Do**
- [ ] Write tests?
- [x] Add readme with demo

### DIFF
--- a/packages/plugins/firebase-auth/README.md
+++ b/packages/plugins/firebase-auth/README.md
@@ -1,0 +1,3 @@
+## `@envelop/firebase-auth`
+
+TODO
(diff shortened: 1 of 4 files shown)


YassinEldeeb:
Hey @saihaj 👋🏼

This looks pretty cool, I think the only thing missing is tests right? would love to know more about the current state of this draft 👀

PR #688, opened 2021-09-15
feat: use apollo server v3 and set default page to graphql playground

### LINKED ISSUE #603: Fix examples with Apollo-Server v3

saihaj:
Update example to use Apollo v3 and also add graphql playground as default instead of Apollo Studio. I found it confusing that it takes to studio cause I just want to be able to query and see it work.

Fixes #603


dotansimha (approved):
looks great, can you please rebase? :)

saihaj:
Should be good now @dotansimha

PR #711, opened 2021-09-22
Allow to set ttl=0 to disable caching, and use ttlPerType to maintain…

dotansimha:


n1ru4l (approved):

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -832,4 +917,212 @@ describe('useResponseCache', () => {
...
+            id
+            text
+          }
+        }
+      }
+    `;
+
+    await testInstance.execute(query);
+    await testInstance.execute(query);
+    expect(spy).toHaveBeenCalledTimes(2);
```
kamilkisiela, line 994:
```suggestion
    const result1 = await testInstance.execute(query);
    const result2 = await testInstance.execute(query);
    expect(result1.errors).not.toBeDefined();
    expect(result2.errors).not.toBeDefined();
    expect(spy).toHaveBeenCalledTimes(2);
```

PR #725, opened 2021-09-26
Masked error plugin: Configure custom message used to mask errors

dthyresson:
## Description

When using the maskedErrors plugin with RedwoodJS, we noticed that when we encounter errors raised within Context Building -- for example when we fetch the authentication headers or try to resolve the currentUser as part of the authentication flow -- then the plugin's formatError is called. And therefore, we message returned is the default "Unexpected error.".

Note: the maskedError plugin is the last plugin added (that is our custom auth plugins are registered before maskedErrors -- in case perhaps that is the issue).

It would be nice if those messages were consistent -- that is either:

* always use the custom formatError
* allow maskedError to defined a custom error message and use that when formatting.

Fixes # https://github.com/dotansimha/envelop/issues/724

This PR adds a custom message config -- it does not aim to address the formatError used in context building.

## Type of change

## How Has This Been Tested?

- [ ] Adds tests for existing with "with a custom error message" scenario

**Test Environment**:

- OS:
- `@envelop/...`:
- NodeJS:

## Checklist:

## Further comments

Perhaps addressing why a custom formatError isn't used when errors happen in context building or init is better, but it is still useful to expose a custom message to use for flexible messaging.

### DIFF
--- a/.changeset/few-horses-exist.md
+++ b/.changeset/few-horses-exist.md
@@ -1,0 +1,5 @@
+---
+'@envelop/core': minor
+---
+
+Adds a custom message option used when masking errors
--- a/packages/core/docs/use-masked-errors.md
+++ b/packages/core/docs/use-masked-errors.md
@@ -36,3 +36,27 @@ const getEnveloped = envelop({
   plugins: [useSchema(schema), useMaskedErrors()],
 });
 ```
+
+You may customize the default error message `Unexpected error.` with your own `errorMessage`:
+
+```ts
+const getEnveloped = envelop({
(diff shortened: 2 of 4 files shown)


dotansimha (approved):
❤️

dotansimha:
Thanks @dthyresson , can you please just add a changset? ( I think `minor` bump is fine in this case)

dthyresson:
> Thanks @dthyresson , can you please just add a changset? ( I think `minor` bump is fine in this case)

Sure will do. I have not done that before but will figure out how.

> Use the changeset cli for creating a changeset for the PR

dthyresson:
I added a changes (I think) with a minor bump and also a few documentation notes for setting the custom options (message and formatter).

PR #745, opened 2021-09-29
useErrorHandler: pass in the context to the error handlers

dios-david:
## Description

Related https://github.com/dotansimha/envelop/issues/744

For detailed error reporting it would be really useful to access the context in the error handlers.

I'm not 100% convienced this is the best approach as I had to do lot of juggling with `ContextType`. If there is a better solution I'm happy to update the PR.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## How Has This Been Tested?

- [x] `packages/core/src/plugins/use-error-handler.ts` - There were no tests for this plugin so I implemented a basic one to demonstrate how the plugin should work.

**Test Environment**:

- OS: macOS 11.1
- `@envelop/...`: latest
- NodeJS: v14.15.4

## Checklist:

### DIFF
--- a/packages/core/docs/use-error-handler.md
+++ b/packages/core/docs/use-error-handler.md
@@ -8,7 +8,7 @@ import { buildSchema } from 'graphql';
 
 const getEnveloped = envelop({
   plugins: [
-    useErrorHandler((errors, context) => {
+    useErrorHandler((errors, args) => {
       // This callback is called once, containing all GraphQLError emitted during execution phase
     }),
     // ... other plugins ...
--- a/packages/core/docs/use-error-handler.md
+++ b/packages/core/docs/use-error-handler.md
@@ -8,7 +8,7 @@ import { buildSchema } from 'graphql';
 
 const getEnveloped = envelop({
   plugins: [
-    useErrorHandler(error => {
+    useErrorHandler((error, context) => {
       // This callback is called per each GraphQLError emitted during execution phase
     }),
     // ... other plugins ...
--- a/packages/core/docs/use-error-handler.md
+++ b/packages/core/docs/use-error-handler.md
@@ -1,6 +1,6 @@
 #### `useErrorHandler`
 
-This plugin triggers a custom function every time execution encounter an error.
+This plugin triggers a custom function when execution encounters an error.
 
 ```ts
 import { envelop, useErrorHandler } from '@envelop/core';
@@ -8,12 +8,10 @@ import { buildSchema } from 'graphql';
 
 const getEnveloped = envelop({
   plugins: [
-    useErrorHandler((error, context) => {
-      // This callback is called per each GraphQLError emitted during execution phase
+    useErrorHandler((errors, context) => {
+      // This callback is called once, containing all GraphQLError emitted during execution phase
     }),
     // ... other plugins ...
   ],
 });
 ```
-
-> Note: every error is being triggered on it's own. So an execution results will multiple error will yield multiple calls.
--- a/packages/core/src/orchestrator.ts
+++ b/packages/core/src/orchestrator.ts
@@ -26,6 +26,7 @@ import {
   isAsyncIterable,
   OnContextErrorHandler,
   SubscribeErrorHook,
+  DefaultContext,
 } from '@envelop/types';
 import {
   DocumentNode,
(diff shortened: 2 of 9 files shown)


packages/core/test/plugins/use-error-handler.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    },
+  });
+
+  it('should invoke error handler when error happens during execution', async () => {
+    const mockHandler = jest.fn();
+    const testInstance = createTestkit([useErrorHandler(mockHandler)], schema);
+    await testInstance.execute(`query { foo }`, {}, { foo: 'bar' });
+
+    expect(mockHandler).toHaveBeenCalledWith([testError], expect.objectContaining({ foo: 'bar' }));
+  });
```
n1ru4l, line 37: Could you also add a test case for when an error is thrown within `useExtendContext`?

n1ru4l: Nevermind, I was confusing this with useMaskedErrors. All good :)

packages/core/src/plugins/use-error-handler.ts
```diff
@@ -1,19 +1,24 @@
...
+  <ContextType>(errorHandler: ErrorHandler) =>
+  ({ context, result }: ErrorHandlerCallback<ContextType>) => {
     if (result.errors?.length) {
-      errorHandler(result.errors);
+      errorHandler(result.errors, context);
     }
   };
 
-export const useErrorHandler = (errorHandler: ErrorHandler): Plugin => ({
+export const useErrorHandler = <ContextType>(errorHandler: ErrorHandler): Plugin<ContextType> => ({
```
n1ru4l, line 19:
```suggestion
export const useErrorHandler = <ContextType = DefaultContext>(errorHandler: ErrorHandler): Plugin<ContextType> => ({
```

n1ru4l (approved):

n1ru4l (commented):
I am wondering whether we should pass in the whole execution arguments instead of only the context? I could imagine people asking about adding that later on?

packages/core/docs/use-error-handler.md
```diff
@@ -1,19 +1,17 @@
 #### `useErrorHandler`
 
-This plugin triggers a custom function every time execution encounter an error.
```
dios-david, line 3: Btw this part didn't reflect the reality, that's why I updated it. I assume the behaviour was changed at some point without updating the docs.

packages/core/test/execute.spec.ts
```diff
@@ -31,6 +31,7 @@ describe('execute', () => {
     expect(spiedPlugin.spies.afterResolver).toHaveBeenCalledTimes(3);
     expect(spiedPlugin.spies.afterExecute).toHaveBeenCalledTimes(1);
     expect(spiedPlugin.spies.afterExecute).toHaveBeenCalledWith({
+      args: expect.any(Object),
```
dios-david, line 34: I think in long term it would be useful to create custom matchers for the main types like execution arguments, then we could just do `args: expect.toBeExecutionArguments` or something like that

dios-david:
@n1ru4l Great idea, I changed this to pass in the execution arguments instead. I think this is way better as other plugins are usually exposing the execution args, not just the context.

packages/core/src/orchestrator.ts
```diff
@@ -56,7 +57,9 @@ export type EnvelopOrchestrator<
   getCurrentSchema: () => Maybe<GraphQLSchema>;
 };
 
-export function createEnvelopOrchestrator<PluginsContext = any>(plugins: Plugin[]): EnvelopOrchestrator<any, PluginsContext> {
+export function createEnvelopOrchestrator<PluginsContext extends DefaultContext>(
```
dios-david, line 60: I spent some time to understand the hook typings, I think my changes are correct now.
I changed the signature of `createEnvelopOrchestrator<PluginsContext = any>` to `createEnvelopOrchestrator<PluginsContext extends DefaultContext>` which makes the typings more clear.
If I understand this right, the `PluginsContext` reflects the type of the `contextValue` used in the current envelop instance, so it made sense to make it extend `DefaultContext` instead of just `any`.

packages/core/test/execute.spec.ts
dotansimha: Agree!

n1ru4l (approved):

PR #758, opened 2021-10-02
feat: add enable option for plugins

saihaj:
Add `enable` option to plugin which can be used to conditionally enable plugins.

### DIFF
--- a/packages/types/src/utils.ts
+++ b/packages/types/src/utils.ts
@@ -24,6 +24,7 @@ export type TuplifyUnion<T, L = LastOf<T>, N = [T] extends [never] ? true : fals
   : Push<TuplifyUnion<Exclude<T, L>>, L>;
 
 export type Unarray<T> = T extends Array<infer U> ? U : T;
+export type NullableArray<T> = T extends Array<infer U> ? (U | null | undefined)[] : T[];
 
 export type ArbitraryObject = Record<string | number | symbol, any>;
 export type PromiseOrValue<T> = T | Promise<T>;
--- a/packages/types/src/utils.ts
+++ b/packages/types/src/utils.ts
@@ -24,7 +24,6 @@ export type TuplifyUnion<T, L = LastOf<T>, N = [T] extends [never] ? true : fals
   : Push<TuplifyUnion<Exclude<T, L>>, L>;
 
 export type Unarray<T> = T extends Array<infer U> ? U : T;
-export type NullableArray<T> = T extends Array<infer U> ? (U | null | undefined)[] : T[];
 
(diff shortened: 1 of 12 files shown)


PabloSzx:
isn't it better to allow "null | undefined | boolean" in the list of plugins and when envelop reads the list it filters them? that's a very common usage pattern for plugins

for example, this is what rollup does, (and graphql-ez too)
image


this allows for:

```ts
[process.env.ENABLE_TRACING ? useTracing() : false, useSchema(schema)]
```

or 
```ts
[process.env.ENABLE_TRACING && useTracing(), useSchema(schema)]
```

or equivalents

saihaj:
>  ```ts
> [process.env.ENABLE_TRACING && useTracing(), useSchema(schema)]
> ```

This looks clean we can do that too.

dotansimha:
Thanks @saihaj ! 
I was thinking that maybe we can export a helper like: 

```
import { enableIf } from '@envelop/core';

  plugins: [
    enableIf(isDev, somePlugin)
  ]
});
```

Then, the core can just filter the `null` plugins. This way this mechanism will support all plugins, without the need to modify all of them.

What do you think?

saihaj:
yea that is better I will update this

packages/types/src/utils.ts
```diff
@@ -24,6 +24,7 @@ export type TuplifyUnion<T, L = LastOf<T>, N = [T] extends [never] ? true : fals
   : Push<TuplifyUnion<Exclude<T, L>>, L>;
 
 export type Unarray<T> = T extends Array<infer U> ? U : T;
+export type NullableArray<T> = T extends Array<infer U> ? (U | null | undefined)[] : T[];
```
dotansimha: Doing `NullableArray<PluginType>` will result in `PluginType[]` without `null`, is that intended?

saihaj: Yes this is just to `null` values in an array so if it just a primitive it shouldn’t do anything. So it should just return `T`. I think better is `Nullable<T>` will fix it

saihaj: What do you think if you don’t pass  an array should it return it back as is Ike null the values?

dotansimha: It can just be: `type NullableArray<T> = Array<T | null>`?

dotansimha:
@saihaj can you please review my changes?
I created a branded nomial value instead of `null` because it felt a bit weird to allow external API to use `null` as value. I used a private nominal value and added a helper function based on TS type guard. 

Also, pushed a few minor fixes to fix circular imports and issues with missing peerDependencies.

saihaj:
Lgtm. That’s very smart 😎

PR #762, opened 2021-10-03
useResponseCache - Prevents unwanted caching of introspection queries and error/no data results

### LINKED ISSUE #761: useResponseCache - Unwanted caching of introspection queries and when errors or no data
While testing the Redis-backed cache in real-world case using a Netlify serverless function (see: https://github.com/dthyresson/envelop-redis-cache-example), I noticed that:

1. Introspection queries are cached
2. When query or mutation has an error, the error result is cached
3. When data is null, null data results are cached

**To Reproduce**

For 1, while in development and the schema is changing and the ttl for the cache is long (or infinite), when using the playground or any dev query tool like Insomnia, the introspection query is cached. This results in stale schema and docs and one is forced to flush the Redis cache since no mutation or ops result key is easily used to purge it.

This could be an option.

For 2, when the connection to Redis failed -- or when in dev and making a mistake in a resolver such that an error is thrown, noticed that for the query or mutation, the result with the errors was cached. Again, had to wait for til to expire, flash cache or fix a mutation such that it might invalidate.

For 3, this is similar to 2. When returning null for some reason, no data is returned for a query. This was cached and thus the ops result key lingered. Seems reasonable not to cache "no data returned".

**Environment:**

- OS: OSX/M1
- `@envelop/response-cache` and see with `@envelop/response-cache-redis` but happens with LRU cache, too
- NodeJS:

dthyresson:
## Description

See: https://github.com/dotansimha/envelop/issues/761

While testing the Redis-backed cache in real-world case using a Netlify serverless function (see: https://github.com/dthyresson/envelop-redis-cache-example), I noticed that:

* Introspection queries are cached
* When query or mutation has an error, the error result is cached
* When data is null, null data results are cached

Fixes #761 

## Type of change

## How Has This Been Tested?

Tests have been added for errors, no data, and optional introspection caching:

- [x] Should not cache an introspection query by default
- [x] Should not cache an introspection query if cacheIntrospections is false
- [x] should not cache query when data is null
- [x] should not cache query when errors
- [x] should not purge cache on mutation when error'

**Test Environment**:

- OS: OSX/M1
- `@envelop/response-cache.`
- NodeJS: v14.17.1

## Checklist:

- [x] My changes generate no new warnings (have added console warns)

## Further comments

* I am not sure about the "no data" or null data case. Perhaps there are good reasons to cache that result.
* I was unsure of the way to detect Introspection queries, but checking if the result is an inspection document seemed tp work
* I did trying to see if the query was an inspection operation in the onParse, but that generated more warnings later if that there was no document
* Last, the test for the inspection query is one that compares the results of two introspection queries -- before and after a schema change. I used two testKits with the same cache instance. Ideally, I would have liked to use the same testKit but swap out the schema for an onSchemaChange event but could not figure that out


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -168,6 +176,10 @@ export function useResponseCache({
       const identifier = new Map<string, CacheEntityRecord>();
       const types = new Set<string>();
 
+      if (!cacheIntrospections && isIntrospectionDocument(ctx.args.document)) {
```
dotansimha: Maybe this could go to the beginning of the `onExecute` function? just to avoid creating the Map/Set objects

dthyresson: This is so strange, I don't have the code in onParse anymore but in

```ts
    async onExecute(ctx) {
      const identifier = new Map<string, CacheEntityRecord>();
      const types = new Set<string>();

      if (!cacheIntrospections && isIntrospectionDocument(ctx.args.document)) {
        return;
      }
````

But, it is late and I may not be reading my code properly :)

**Comment by dthyresson** [comment]
Based on feedback, I have refactored to include a custom `shouldCacheResult` function with a default behavior that matches the above.

Note: the `cacheIntrospections` boolean is still present, though including it could be confusing. I could see the argument to remove it and let all that logic live in `shouldCacheResult`.

File: packages/plugins/response-cache/src/plugin.ts (hunk 1 above)
Comment (dthyresson) [reply packages/plugins/response-cache/src/plugin.ts]: All logic for if should cache result is not in a customizable function: shouldCacheResult.

File: packages/plugins/response-cache/src/plugin.ts [hunk 2]
```diff
@@ -123,6 +142,45 @@ export const defaultBuildResponseCacheKey: BuildResponseCacheKeyFunction = param
...
+    }
+  } else {
+    if (params.result) {
+      if (params.result?.errors) {
+        // eslint-disable-next-line no-console
+        console.warn('[useResponseCache] Failed to cache or invalidate due to errors');
+        return false;
+      }
+
+      if (params.result?.data === null || params.result?.data === undefined) {
```
n1ru4l: I think if `result.errors` is present in `onExecuteDone` this implies that `data` is either `null` or an `object`. I think `data` can never be `undefined` or `null` here as the previous check in line 167 already implies that data must be a `object` and is never `null` or `undefined`. Thus this check might be unnecessary.

n1ru4l:
As I am looking at this code I was thinking whether introspection could simply be disabled by setting a ttl for `Query.__schema` to `0`.

```ts
  plugins: [
    // ... other plugins ...
    useResponseCache({
      ttlPerSchemaCoordinate: {
        "Query.__schema": 0
      }
    }),
  ],
});
```

or even by setting the `__Schema` type ttl to `0`.

```ts
  plugins: [
    // ... other plugins ...
    useResponseCache({
      ttlPerType: {
        __Schema: 0
      }
    }),
  ],
});
```

We could make this the default value and thus reduce the complexity of the plugin (more config options, that clash).

I agree that we still need a way of determining whether a single operation result should be cached.

I like the approach of `ShouldCacheResultFunction`

```ts
export type ShouldCacheResultFunction = (params: {
  cacheIntrospections?: Boolean;
  documentNode?: DocumentNode;
  result?: ExecutionResult;
}) => Boolean;
```

If can simply remove the `cacheIntrospections` param. Also is there a reason why `documentNode` is optional? I don't think we need this. I saw that `shouldCacheResult` is used in multiple places. Only the one in `onExecuteDone` is really necessary.

@dthyresson please let me know what you think.

dotansimha:
> As I am looking at this code I was thinking whether introspection could simply be disabled by setting a ttl for `Query.__schema` to `0`.
> 
> ```ts
> const getEnveloped = envelop({
>   plugins: [
>     // ... other plugins ...
>     useResponseCache({
>       ttlPerSchemaCoordinate: {
>         "Query.__schema": 0
>       }
>     }),
>   ],
> });
> ```
> 
> or even by setting the `__Schema` type ttl to `0`.
> 
> ```ts
> const getEnveloped = envelop({
>   plugins: [
>     // ... other plugins ...
>     useResponseCache({
>       ttlPerType: {
>         __Schema: 0
>       }
>     }),
>   ],
> });
> ```
> 
> We could make this the default value and thus reduce the complexity of the plugin (more config options, that clash).
> 
> I agree that we still need a way of determining whether a single operation result should be cached.
> 
> I like the approach of `ShouldCacheResultFunction`
> 
> ```ts
> export type ShouldCacheResultFunction = (params: {
>   cacheIntrospections?: Boolean;
>   documentNode?: DocumentNode;
>   result?: ExecutionResult;
> }) => Boolean;
> ```
> 
> If can simply remove the `cacheIntrospections` param. Also is there a reason why `documentNode` is optional? I don't think we need this. I saw that `shouldCacheResult` is used in multiple places. Only the one in `onExecuteDone` is really necessary.
> 
> @dthyresson please let me know what you think.

Oh, that's clever! I think it covers the need to disable caching for introspection! 
@n1ru4l maybe you can listen to `onSchemaChange` and invalidate this coordinate as well? This way users won't need to add the TTL at all.

n1ru4l:
@dotansimha I guess the idea is to never cache introspection at all, so maybe we can merge the schema coordinate and object TTL with a default object where those values are set to 0 🤔

dthyresson:
> Also is there a reason why `documentNode` is optional? I don't think we need this. I saw that `shouldCacheResult` is used in multiple places. Only the one in `onExecuteDone` is really necessary.

@n1ru4l The reason I had both `documentNode` and `result` is that `documentNode` helped check if was an introspection query. The `result` is used if you want to prevent results with errors from being cached.

This PR tried to address two cases: optionally cache Introspection (but by default don't) because as you noted, if in Production you may want to cache it for performance (though it is tough to purge) and also don't cache error results.

```ts
/**
 * Default function used to check if the result should be cached.
 * It is exported here for advanced use-cases. E.g. if you want to
 * cache Introspection queries or choose if results with certain error types
 * should be cached.
 *
 * By default, introspection queries are not cached. Neither are any results
 * with errors (unexpected, EnvelopError or GraphQLError) or results will missing data.
 */
export const defaultShouldCacheResult: ShouldCacheResultFunction = (params: {
  cacheIntrospections?: Boolean;
  documentNode?: DocumentNode;
  result?: ExecutionResult;
}): Boolean => {
  if (params.documentNode) {
    if (!params.cacheIntrospections && isIntrospectionDocument(params.documentNode)) {
      console.warn('[useResponseCache] Did not cache IntrospectionQuery');
      return false;
    }
  } else {
    if (params.result) {
      if (params.result?.errors) {
        console.warn('[useResponseCache] Failed to cache or invalidate due to errors');
        return false;
      }

      if (params.result?.data === null || params.result?.data === undefined) {
        console.warn('[useResponseCache] Failed to cache due to missing data');
        return false;
      }
    }
  }
```

If we want to never cache introspections, then I can just keep the `result` and error check. And use the schema coordinate approach.

dthyresson:
Actually, that `ShouldCacheResultFunction` only takes a required `result` makes more sense than adding in a `documentNode`. "should cache ** result**" and hence pass the result :)

packages/plugins/response-cache/src/plugin.ts
n1ru4l: @dthyresson Thoughts in this?

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -133,17 +169,21 @@ export function useResponseCache({
...
   getDocumentStringFromContext = defaultGetDocumentStringFromContext,
+  shouldCacheResult = defaultShouldCacheResult,
   // eslint-disable-next-line dot-notation
   includeExtensionMetadata = typeof process !== 'undefined' ? process.env['NODE_ENV'] === 'development' : false,
 }: UseResponseCacheParameter = {}): Plugin {
   const ignoredTypesMap = new Set<string>(ignoredTypes);
   const schemaCache = new WeakMap<GraphQLSchema, GraphQLSchema>();
 
+  // never cache Introspections
+  ttlPerSchemaCoordinate = { ...ttlPerSchemaCoordinate, 'Query.__schema': 0 };
```
n1ru4l:
```suggestion
  ttlPerSchemaCoordinate = { 'Query.__schema': 0, ...ttlPerSchemaCoordinate };
```

Just in case someone actually wants to introduce TTL for `Query.__schema` - Then he can override the default value of 0 or even enable long-living introspection caching by setting the value to `undefined` 🤔

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -194,6 +235,10 @@ export function useResponseCache({
               return;
             }
 
+            if (!shouldCacheResult({ result })) {
+              return;
+            }
```
n1ru4l: Mutations must never be cached so this does not make so much sense here?

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -123,6 +132,33 @@ export const defaultBuildResponseCacheKey: BuildResponseCacheKeyFunction = param
...
+      // eslint-disable-next-line no-console
+      console.warn('[useResponseCache] Failed to cache or invalidate due to errors');
+      return false;
+    }
+
+    if (params.result?.data === null || params.result?.data === undefined) {
+      // eslint-disable-next-line no-console
+      console.warn('[useResponseCache] Failed to cache due to missing data');
+      return false;
+    }
```
n1ru4l: Aside from the fact that this check might be unnecessary (as it is covered by the `params.result.errors` check - The optional chaining operator here is not required. As there is a  falsy check earlier for `params.result`, the typescript compiler knows that `params.result` is not `null` or `undefined`. This you can simply write `params.result.data === null || params.result.data === undefined`.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -123,6 +132,33 @@ export const defaultBuildResponseCacheKey: BuildResponseCacheKeyFunction = param
...
+ *
+ * It is exported here for advanced use-cases. E.g. if you want to choose if
+ * results with certain error types should be cached.
+ *
+ * By default, results with errors (unexpected, EnvelopError or GraphQLError)
+ * or results with missing data are not cached.
+ */
+export const defaultShouldCacheResult: ShouldCacheResultFunction = (params: { result: ExecutionResult }): Boolean => {
+  if (params.result) {
+    if (params.result?.errors) {
```
n1ru4l:
```suggestion
    if (params.result.errors) {
```

packages/plugins/response-cache/src/plugin.ts
dthyresson: If I remove this, my test for 'Should not purge cache on mutation when error' fails though. Let me see why.

image

dthyresson: The test I have attempts a mutation, but the mutation encounters an error. Maybe the db update didn't succeed or something else.

So, because of that the mutation ... but,  right it should never have been cached. So this a strange failure.

Perhaps my test case is incorrect.

packages/plugins/response-cache/src/plugin.ts
dthyresson: I removed and my tests still pass, so will remove that `result.data` check altogether.

packages/plugins/response-cache/src/plugin.ts
dthyresson: @n1ru4l This is the only item perhaps you can take on to complete this PR. 

Since my test fails if I remove that check, maybe you can see if perhaps my test case isn't correct and thus giving a false result.

Thanks!

n1ru4l: I think we should still purge the cache if errors occur. I am more worried that not doing so will have unexpected behavior and results in people hitting a stale cache. This is currently just an assumption. Real-world usage will hopefully give us a better understanding of what makes the most sense. I adjusted this behavior in https://github.com/dotansimha/envelop/pull/762/commits/a26808ad87e6e405775a668ba7b352cd69d97e49

dthyresson (commented):
Thanks!

n1ru4l:
@dthyresson I am going through some test cases right now - Might ask some more questions!

n1ru4l (approved):
Thank you so much. 🎉

I simplified some test cases and also added one for enabling introspection caching.

packages/plugins/response-cache/README.md
```diff
@@ -191,19 +191,98 @@ const getEnveloped = envelop({
 });
 ```
 
-### Prevent caching of sensitive information
+### Customize if result should be cached
+
+You can define s custom function used to check if the result should be cached.
```
n1ru4l:
```suggestion
You can define a custom function used to check if the result should be cached.
```

packages/plugins/response-cache/README.md
```diff
@@ -191,19 +191,98 @@ const getEnveloped = envelop({
...
+
+This is useful for advanced use-cases. E.g. if you want to
+cache Introspection queries or choose if results with certain error types
+should be cached.
```
n1ru4l:
```suggestion
This is useful for advanced use-cases. E.g. if you want to
cache results with certain error types.
```

packages/plugins/response-cache/README.md
```diff
@@ -191,19 +191,98 @@ const getEnveloped = envelop({
...
+
+By default it used the `defaultShouldCacheResult` function:
```
n1ru4l:
```suggestion
By default, it uses the `defaultShouldCacheResult` function which never caches any query operation execution results that include errors (unexpected, EnvelopError, or GraphQLError).
```

packages/plugins/response-cache/README.md
```diff
@@ -191,19 +191,98 @@ const getEnveloped = envelop({
...
+
+- Introspection queries are not cached.
+- Neither are any results with errors (unexpected, EnvelopError or GraphQLError) or
+- Results will missing data
```
n1ru4l:
```suggestion
```

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -123,6 +136,25 @@ export const defaultBuildResponseCacheKey: BuildResponseCacheKeyFunction = param
...
     .digest('base64');
 
+/**
+ * Default function used to check if the result should be cached.
+ *
+ * It is exported here for advanced use-cases. E.g. if you want to choose if
+ * results with certain error types should be cached.
+ *
+ * By default, results with errors (unexpected, EnvelopError or GraphQLError)
+ * or results with missing data are not cached.
```
n1ru4l:
```suggestion
 * By default, results with errors (unexpected, EnvelopError, or GraphQLError) are not cached.
```

PR #854, opened 2021-10-21
doc: enableif utility

saihaj:
document #758


saihaj:
Not sure how benchmarks got affected

PR #864, opened 2021-10-23
Improve CI

ardatan:
- Create a tsconfig.build.json file to exclude tests and stuff so build process doesn't do type checking on test files but we still use `tsconfig.json` to do type checking on test files `yarn ts:check` and I've added a workflow for that.
- Extract ESM Test, Lint, Build and Unit Testing into different workflows
- Use babel typescript instead of ts jest for performance
- CI workflow is ready to have tests for multiple GraphQL versions. GraphQL v14 was failing, that's why I couldn't enable it. After merging this, I can work on fixing the issues for v14 and v16.
- Cache node_modules instead of yarn cache folder


.github/workflows/tests.yml
```diff
@@ -0,0 +1,122 @@
...
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Master
+        uses: actions/checkout@v2
+      - name: Use Node
+        uses: actions/setup-node@master
+        with:
+          node-version: '16.8.0'
```
PabloSzx: why 16.8.0 in specific when 16.12.0 is already out?

ardatan: I wasn't sure if this bug has been fixed;
https://github.com/nodejs/node/issues/40030

ardatan:
@dotansimha Let's merge asap so we can extend the matrix with v14 and v16

PR #870, opened 2021-10-24
Avoid nested imports from graphql-js

ardatan:
Remove nested imports from `graphql-js` like `graphql/*` because the deeper exports are not the same in v16

### DIFF
--- a/packages/plugins/extended-validation/src/rules/one-of.ts
+++ b/packages/plugins/extended-validation/src/rules/one-of.ts
@@ -1,5 +1,5 @@
 import { ArgumentNode, GraphQLError, GraphQLInputObjectType, GraphQLInputType, isListType, ValidationContext } from 'graphql';
-import { getArgumentValues } from 'graphql/execution/values.js';
+import { getArgumentValues } from '@graphql-tools/utils';
 import { ExtendedValidationRule, getDirectiveFromAstNode, unwrapType } from '../common';
 
 export const ONE_OF_DIRECTIVE_SDL = /* GraphQL */ `
@@ -18,7 +18,7 @@ export const OneOfInputObjectsRule: ExtendedValidationRule = (validationContext,
           return;
         }
 
-        const values = getArgumentValues(fieldType, node, executionArgs.variableValues);
+        const values = getArgumentValues(fieldType, node, executionArgs.variableValues || undefined);
 
         if (fieldType) {
           const isOneOfFieldType =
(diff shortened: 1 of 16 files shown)


packages/plugins/extended-validation/src/rules/one-of.ts
```diff
@@ -1,5 +1,5 @@
 import { ArgumentNode, GraphQLError, GraphQLInputObjectType, GraphQLInputType, isListType, ValidationContext } from 'graphql';
-import { getArgumentValues } from 'graphql/execution/values.js';
+import { getArgumentValues } from '@graphql-tools/utils';
```
ardatan, line 2: Not sure if we have zero-dependency policy.
`getArgumentValues` is not something directly exported so I imported it from `@graphql-tools/utils`

packages/types/src/async-utils.ts
```diff
@@ -5,16 +5,19 @@ import {
...
  * Source: https://github.com/graphql/graphql-js/blob/main/src/jsutils/isAsyncIterable.ts
  */
-export function isAsyncIterable(maybeAsyncIterable: any): maybeAsyncIterable is AsyncIterableIterator<ExecutionResult> {
-  return typeof maybeAsyncIterable?.[Symbol.asyncIterator] === 'function';
+export function isAsyncIterable(maybeAsyncIterable: any): maybeAsyncIterable is AsyncIterable<any> {
+  return (
+    maybeAsyncIterable != null &&
+    typeof maybeAsyncIterable === 'object' &&
+    typeof maybeAsyncIterable[Symbol.asyncIterator] === 'function'
+  );
```
n1ru4l, line 20: Why this version? What is the functional difference?

packages/plugins/resource-limitations/src/index.ts
```diff
@@ -14,7 +14,7 @@ import {
   GraphQLType,
   isScalarType,
 } from 'graphql';
-import { getArgumentValues } from 'graphql/execution/values.js';
+import { getArgumentValues } from '@graphql-tools/utils';
```
n1ru4l, line 17: Technically we only have a zero dependency policy for the core. Ideally we avoid deps wherever possible. it is a bummer that we have to add a dependency for one function -.-

ardatan: So we need to copy it from graphql-js?

packages/types/src/async-utils.ts
ardatan: `maybeAsyncIterable` might be a primitive value or null(which is an `object`) or an `AsyncIterable` object with the iterator inside.
This is actually something I noticed while reading the code. I can revert it no worries but at least `is AsyncIterable<ExecutionResult>` should be `AsyncIterable<any>` because this doesn't check the content of `AsyncIterable`. For example it doesn't act correctly while using inside `createSubscribe`.

PR #871, opened 2021-10-24
feat(graphql-jit): subscription support

ardatan:


saihaj (changes requested):
we should add test for this :) similar to how we have for `execute`

ardatan:
Yes we need to add "subscribe" to testkit

ardatan:
Ok no subscribe is needed! It worked. PR is ready @dotansimha

dotansimha (approved):

saihaj (approved):
Awesome!

PR #893, opened 2021-10-26
fix: handle introspection types per field

### LINKED ISSUE #892: operation-field-permissions: possibility to bypass the authorization
With the current implementation of skipping introspection queries, it is possible to bypass authorization by including `__schema` query in any other query

**To Reproduce**

add `__schema { __typename }` to any query to bypass authorization, like in the test:
```
query {
  __schema {
    __typename
  }
  greetings
}
```

dimatill:
## Description

Perform authorization for queries that contain `__schema` query.
Skip authorization for Introspection Types and nested fields.

Fixes #892 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/plugins/extended-validation/src/plugin.ts
+++ b/packages/plugins/extended-validation/src/plugin.ts
@@ -48,10 +48,6 @@ export const useExtendedValidation = (options: {
       const visitor = visitInParallel(rules.map(rule => rule(validationContext, args)));
       visit(args.document, visitWithTypeInfo(typeInfo, visitor));
 
-      for (const rule of rules) {
-        rule(validationContext, args);
-      }
-
       if (errors.length > 0) {
         let result: ExecutionResult = {
           data: null,
--- a/packages/plugins/operation-field-permissions/src/index.ts
+++ b/packages/plugins/operation-field-permissions/src/index.ts
@@ -1,4 +1,4 @@
-import { Plugin, useExtendContext, isIntrospectionDocument } from '@envelop/core';
+import { Plugin, useExtendContext } from '@envelop/core';
 import { ExtendedValidationRule, useExtendedValidation } from '@envelop/extended-validation';
 import {
   GraphQLType,
@@ -10,6 +10,7 @@ import {
   GraphQLObjectType,
   isObjectType,
   isInterfaceType,
+  isIntrospectionType,
 } from 'graphql';
 
 type PromiseOrValue<T> = T | Promise<T>;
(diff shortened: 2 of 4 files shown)


packages/plugins/extended-validation/src/plugin.ts
```diff
@@ -48,10 +48,6 @@ export const useExtendedValidation = (options: {
       const visitor = visitInParallel(rules.map(rule => rule(validationContext, args)));
       visit(args.document, visitWithTypeInfo(typeInfo, visitor));
 
-      for (const rule of rules) {
-        rule(validationContext, args);
-      }
```
dimatill, line 53: looks like it's not needed because of `rules.map(rule => rule(validationContext, args))` few lines above

But I'm not sure if this change should go with a separate PR. Please let me know if so

dimatill:
looks like some checks are failed, I'll look into it later today

dimatill:
> looks like some checks are failed, I'll look into it later today

rebased the branch and now all looks good :)

packages/plugins/operation-field-permissions/src/index.ts
```diff
@@ -79,10 +76,23 @@ const OperationScopeRule =
...
+          const wrappedType = getWrappedType(type);
+
+          if (isIntrospectionType(wrappedType)) {
+            return false;
+          }
+        }
+
         const parentType = context.getParentType();
         if (parentType) {
           const wrappedType = getWrappedType(parentType);
```
dimatill: as I can see it's not necessary to call `getWrappedType` here because `parentType` logically can't be a List nor NonNull type, and type definitions say that it is `GraphQLCompositeType`
```
export type GraphQLCompositeType =
  | GraphQLObjectType
  | GraphQLInterfaceType
  | GraphQLUnionType;
```

we can discuss it and if you agree then I can remove it in this PR or as a separate one

dotansimha (approved):
LGTM. 
Waiting for @n1ru4l 's review.

packages/plugins/extended-validation/src/plugin.ts
n1ru4l: @dotansimha Ylu added this any comments? To me it also seems like we can drop this

packages/plugins/operation-field-permissions/src/index.ts
n1ru4l: We can remove it! Please do it 😇

n1ru4l (commented):
Thank you for this! Please apply the other improvements and @dotansimha please check upon my comment, then we are free to merge 🎉

dimatill:
> Thank you for this! Please apply the other improvements and @dotansimha please check upon my comment, then we are free to merge 🎉

Will do it!

packages/plugins/extended-validation/src/plugin.ts
dotansimha: Oh, seems like a duplication? I guess it can be removed

packages/plugins/operation-field-permissions/src/index.ts
dotansimha: Agree!

dimatill:
> > Thank you for this! Please apply the other improvements and @dotansimha please check upon my comment, then we are free to merge 🎉
> 
> Will do it!

Pushed it up

n1ru4l (approved):

PR #915, opened 2021-11-01
Add `onMissingMatch` option to persisted operations plugin

santino:
## Description

This PR adds a new option to the persisted operations plugin.
This is a callback to allow developers to handle events of missing operation Id/s in store/s.
With this callback, developers can log custom events so that missing matches are kept under control and analysed.

Keep in mind that a missing match is not necessarily an error.
It would be an error only when `onlyPersisted` is true, otherwise, a GraphQL server implementation might even involve a double-way communication so that if an operation Id is not found the client sends that to the server in a second request.

Hence the logging is relevant also in addition to errors handling scenarios.

## Type of change

## Checklist:

### DIFF
--- a/.changeset/flat-zoos-float.md
+++ b/.changeset/flat-zoos-float.md
@@ -1,0 +1,5 @@
+---
+'@envelop/persisted-operations': minor
+---
+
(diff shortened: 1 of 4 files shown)


dotansimha (approved):
Amazing. Thanks @santino !

PR #947, opened 2021-11-08
useSentry: `skipError` option

### LINKED ISSUE #925: useSentry: Option to exclude errors from reporting
## Feature request

There's no way to exclude errors from being reported. Ideally there could be a configuration filter option to opt-out.

Something like:
```js
useSentry({
  skipError(error) { return error instanceof MyCustomError; }
});
```

I'd suggest renaming `skip` to `skipOperation` if we go with `skipError`.

## Bundling bug

I tried to extend my custom errors from [EnvelopError](https://github.com/dotansimha/envelop/blob/dc2ec27dc0ca783f37bcff2a8bba07dd94b8fb7f/packages/core/src/plugins/use-masked-errors.ts#L6-L10) as a workaround, though I was stumped as to why this didn't work until I realised for some reason it's actually being bundled into the Sentry plugin, https://unpkg.com/browse/@envelop/sentry@1.3.0/index.mjs so fails the [instanceof](https://github.com/dotansimha/envelop/blob/main/packages/plugins/sentry/src/index.ts#L167) check, when imported from `@envelop/core`.

dburles:
Adds a new `skipError` option, which allows users to skip certain errors. It's useful in the case where a user has defined custom error types, such as `ValidationError` which may be used to validate resolver arguments.

I've also renamed `skip` to `skipOperation`, this is a breaking change.

Fixes #925

### DIFF
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -60,13 +60,13 @@ export type SentryPluginOptions = {
    */
   operationName?: (args: ExecutionArgs) => string;
   /**
-   * Indicates whether or not to ignore the entire Sentry flow for given GraphQL operation
+   * Indicates whether or not to skip the entire Sentry flow for given GraphQL operation
    */
-  ignoreOperation?: (args: ExecutionArgs) => boolean;
+  skipOperation?: (args: ExecutionArgs) => boolean;
   /**
-   * Indicates whether or not to ignore Sentry exception reporting for a given error,
+   * Indicates whether or not to skip Sentry exception reporting for a given error,
    */
-  ignoreError?: (args: Error) => boolean;
+  skipError?: (args: Error) => boolean;
 };
 
 export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
@@ -80,12 +80,12 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
   const includeRawResult = pick('includeRawResult', false);
   const includeExecuteVariables = pick('includeExecuteVariables', false);
   const renameTransaction = pick('renameTransaction', false);
-  const ignoreOperation = pick('ignoreOperation', () => false);
-  const ignoreError = pick('ignoreError', () => false);
+  const skipOperation = pick('skipOperation', () => false);
+  const skipError = pick('skipError', () => false);
 
   return {
     onExecute({ args }) {
-      if (ignoreOperation(args)) {
+      if (skipOperation(args)) {
         return;
       }
 
@@ -168,7 +168,7 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
                   childSpan.setData('result', result);
                 }
 
-                if (result instanceof Error && !(result instanceof EnvelopError) && !ignoreError(result)) {
+                if (result instanceof Error && !(result instanceof EnvelopError) && !skipError(result)) {
                   const errorPath = responsePathAsArray(info.path).join(' > ');
 
                   Sentry.captureException(result, {
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -69,6 +69,10 @@ export type SentryPluginOptions = {
   skipError?: (args: Error) => boolean;
 };
 
+export function defaultSkipError(error: Error): boolean {
+  return !(error instanceof EnvelopError);
+}
+
 export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
   function pick<K extends keyof SentryPluginOptions>(key: K, defaultValue: NonNullable<SentryPluginOptions[K]>) {
     return options[key] ?? defaultValue;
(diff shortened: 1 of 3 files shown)


packages/plugins/sentry/src/index.ts
```diff
@@ -164,7 +169,7 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
                   childSpan.setData('result', result);
                 }
 
-                if (result instanceof Error && !(result instanceof EnvelopError)) {
+                if (result instanceof Error && !(result instanceof EnvelopError) && !skipError(result)) {
```
n1ru4l: shouldn't the default value for `skipError` be `(result) => result instanceof EnvelopError`?

packages/plugins/sentry/src/index.ts
```diff
@@ -63,7 +63,11 @@ export type SentryPluginOptions = {
   /**
    * Indicates whether or not to skip the entire Sentry flow for given GraphQL operation
    */
-  skip?: (args: ExecutionArgs) => boolean;
+  skipOperation?: (args: ExecutionArgs) => boolean;
+  /**
+   * Indicates whether or not to skip the entire Sentry flow for given error
+   */
+  skipError?: (args: Error) => boolean;
```
n1ru4l, line 71: Technically the error should always be a `GraphQLError`. In which scenario would it "only" be an `Error`?

dburles:
Perhaps `ignore` may be a better name over `skip` when both errors and operations are taken into account.

packages/plugins/sentry/src/index.ts
dburles: That is a good idea.

packages/plugins/sentry/src/index.ts
dburles: Doesn't `GraphQLError` extend `Error`? I could be mistaken. This PR is kind of a sketch to get something in play.

dburles: Happy to take your lead on this as I don't use TypeScript

packages/plugins/sentry/src/index.ts
dburles: Actually, I don't think that will work as there is internal use of `EnvelopError` which should be ignored regardless of what users have defined with `skipError`.

n1ru4l: If people wanna customize the behavior and want to report certain `EnvelopError` instances they need to change the behavior. I think it is totally fine that people need to add this check manually in case they want to customize stuff. We can even export a `defaultSkipError` function that is `(result) => result instanceof EnvelopError`, so people can re-use it.

packages/plugins/sentry/src/index.ts
n1ru4l: Let's keep it `Error` for now, I am not 100% sure, we can still refine it later.

n1ru4l:
We already use the term `skip` in a lot of the plugins, so I would like to actually keep that over `ignore` and keep it familiar.

packages/plugins/sentry/src/index.ts
dburles: I don't really understand what `EnvelopError` instances users would care to report I assume there is a good reason to skip sending them which is why it's in place. Do you have an example scenario in mind? it's not totally clear to me.

For my purposes I wouldn't need to change that behaviour. I just want to skip errors that I've defined.

Maybe if you can elaborate on the idea though that might help me understand what you're getting at a bit better.

n1ru4l: `EnvelopError` is used by the `useMaskedErrors` plugin (https://github.com/dotansimha/envelop/blob/83c27fbceab0dd83377cc5cf344e26292facd656/packages/core/src/plugins/use-masked-errors.ts#L15-L23).  Any error that does not extend `EnvelopError` will be replaced with a GraphQLError with the message `Unexpected error.`.

There might be cases, where people want to report `EnvelopError`s, thus we should keep the flexibility for users to opt into fully controlling what they want to report.

packages/plugins/sentry/src/index.ts
```diff
@@ -77,11 +81,12 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
   const includeRawResult = pick('includeRawResult', false);
   const includeExecuteVariables = pick('includeExecuteVariables', false);
   const renameTransaction = pick('renameTransaction', false);
-  const skip = pick('skip', () => false);
+  const skipOperation = pick('skipOperation', () => false);
+  const skipError = pick('skipError', () => false);
```
n1ru4l: As discussed the default value of `skipError` should be `(err: Error) => err instanceof EnvelopError` and exported from this plugin as `defaultSkipError`

packages/plugins/sentry/src/index.ts
n1ru4l:
```suggestion
                if (result instanceof Error && !skipError(result)) {
```

n1ru4l (changes requested):
As discussed these changes are missing for progressing this further.

n1ru4l:
@dburles Can you add a changeset and a simple test for this feature? Then we can merge and release it 🚀

dotansimha:
Thanks @dburles 
I pushed a commit with the missing `changeset` file, and also reverted `skipOperation` thing. I know it might be better to use `skipOperation` but we can avoid that breaking change and introduce a `minor` release instead of `major`.

PR #955, opened 2021-11-09
Simplify code and add comments in `useExtendedValidation`

maticzav:
This PR adds comments describing how and why code works the way it does. I've also replaced a custom `wrappedType` extracting function with an official one that's non-recursive.


dotansimha (approved):

dotansimha:
Thanks ! ❤️

Issue #976, opened 2021-11-14
New plugin: `useGraphqlExecutor`


dotansimha:
https://github.com/yaacovCR/graphql-executor

saihaj:
with next major https://github.com/n1ru4l/envelop/pull/1487 user will be able to provide a custom `execute` function and not bound to `graphql-js`'s default `execute`. This allows users to customize however they want to.

PR #978, opened 2021-11-15
feat: create graphql-executor plugin

saihaj:
Create GraphQL-executor plugin

Related #976


saihaj:
@yaacovCR thoughts anything I should change on this?

dotansimha (approved):

dotansimha:
@saihaj can you please add it to the website plugin hub? :)

packages/plugins/graphql-executor/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "typings": "dist/index.d.ts",
+  "typescript": {
+    "definition": "dist/index.d.ts"
+  },
+  "scripts": {
+    "test": "jest",
+    "prepack": "bob prepack"
+  },
+  "dependencies": {
+    "graphql-executor": "^0.0.8"
```
n1ru4l: Should this be a peer dependency?

yaacovCR:
Methods are going to change soon with next release, might want to wait a bit for that. Primary method for all operations including subscriptions will just be execute.

also not sure whether it pays in docs to describe how you can use exported execute method or the internal class

also the readme will hopefully change soon as well with the examples and also you reflect actual defer/stream support, inlining text from there is fine, just making a note that might require maintenance

i am currently planning on releasing as v1 with the above changes but may stick with v0.x not sure

packages/plugins/graphql-executor/package.json
dotansimha: I can, but I'm not 100% sure.
I think we can depend on a specific version because this is something users might not install for a different use case?

yaacovCR: You can have multiple versions of graphql executor in a project, so no problem being a regular dependency from that angle (as opposed to graphql js) but maybe you are asking based on other concerns in terms of envelop plug-ins

saihaj:
@yaacovCR so I updated to new version and ran into issue with the types between `TypeExecutionArgs` vs `ExecutionArgs` which I think I can just force it but the other issue it is complaining about different versions `graphql-js`

packages/plugins/graphql-executor/package.json
saihaj: I think having it as a dep is probably easier for now until we get to a stable release similar to how we have `graphql-jit`

packages/plugins/graphql-executor/package.json
```diff
@@ -0,0 +1,73 @@
...
+      },
+      "import": {
+        "types": "./dist/typings/*.d.ts",
+        "default": "./dist/esm/*.js"
+      },
+      "default": {
+        "types": "./dist/typings/*.d.ts",
+        "default": "./dist/esm/*.js"
+      }
+    },
```
n1ru4l, line 43: this section can be removed

n1ru4l, line 43: this section can be removed

packages/plugins/graphql-executor/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@envelop/core": "^1.6.1",
+    "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
+  },
+  "buildOptions": {
+    "input": "./src/index.ts"
+  },
+  "publishConfig": {
+    "directory": "dist",
+    "access": "public"
+  }
```
n1ru4l, line 72: `"type"`: "module"` is missing in this `package.json` - that is why the integrity check fails

packages/plugins/graphql-executor/src/index.ts
```diff
@@ -1,12 +1,12 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+// @ts-nocheck It will work fine it is just the difference in Function signatures.
```
n1ru4l, line 2: I would rather prefer a ts-ignore in the affected line 😅

n1ru4l (changes requested):
See my comments 😇

saihaj:
this gives more customizable/extensible and agnostic solution  https://github.com/n1ru4l/envelop/pull/1487

PR #997, opened 2021-11-19
generic-auth: use extended validation for applying rules

n1ru4l:
Currently, the generic auth plugin applies authorization rules during the execution phase, as the resolvers are called.

I think this has some confusing behavior, as parts of a mutation/query/subscription operation will be executed before yielding an error (that might even bubble up, thus resulting in the whole execution result `data` being `null`).

While building the advanced tutorial examples, this popped up:

```graphql
type Mutation {
    signup(email: String!, password: String!, name: String!): AuthPayload @skipAuth
}

type User {
  id: ID!
  name: String!
}

type AuthPayload {
  token: String
  user: User
}
```

With the current behavior, this allowed executing the operation resolver (`Mutation.signup`), but if you added a selection for the field `token`, you got an unauthenticated error. So the user was created in the database and ultimately you did not receive the response because as the result was constructed the auth rule was applied for the `AuthPayload.token` resolver.

I personally think it makes sense to apply GraphQL based authorization rules via the extended validation phase and then reject either the whole operation document or allow executing the complete operation document.


n1ru4l:
See https://github.com/dotansimha/envelop/pull/998

PR #998, opened 2021-11-19
generic-auth: use extended validation for applying rules

n1ru4l:
Currently, the generic auth plugin applies authorization rules during the execution phase, as the resolvers are called.

I think this has some confusing behavior, as parts of a mutation/query/subscription operation will be executed before yielding an error (that might even bubble up, thus resulting in the whole execution result `data` being `null`).

While building the advanced tutorial examples, this popped up:

```graphql
type Mutation {
  signup(email: String!, password: String!, name: String!): AuthPayload
    @skipAuth
}

type User {
  id: ID!
  name: String!
}

type AuthPayload {
  token: String
  user: User
}
```

With the current behavior, this allowed executing the operation resolver (`Mutation.signup`), but if you added a selection for the field `token`, you got an unauthenticated error. So the user was created in the database and ultimately you did not receive the response because as the result was constructed the auth rule was applied for the `AuthPayload.token` resolver.

I personally think it makes sense to apply GraphQL based authorization rules via the extended validation phase and then reject either the whole operation document or allow executing the complete operation document.


saihaj (approved):

.changeset/six-seals-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+'@envelop/generic-auth': major
+---
+
+wip
```
dotansimha: please change :)

dotansimha (approved):
Looks really nice. Do we have a way to avoid a major / breaking change? 🤔 @n1ru4l

n1ru4l:
@dotansimha We cannot avoid this being a breaking change as we now reject the whole operation instead of allowing partial results :/

packages/plugins/generic-auth/src/index.ts
```diff
@@ -68,85 +87,125 @@ export type GenericAuthPluginOptions<UserType extends {} = {}, ContextType exten
        * resolves the user and inject to authenticated user into the `context`.
        * And checks for `@auth` directives usages to run validation automatically.
        */
-      mode: 'protect-auth-directive';
+      mode: 'protect-single';
```
n1ru4l:
```suggestion
      mode: 'protect-single';
```

@saihaj @dotansimha any better alternatives for the name `protect-single`?

`protect-all` -> all are protected by default

`protect-single` -> more like granular configuration... `protect-manual`? `protect-flexible`?

saihaj: I think `protect-flexible` makes more sense

n1ru4l: I decided upon protect-granular :D

PR #1009, opened 2021-11-22
useSentry: patch for defaultSkipError, which should return true if the error is an EnvelopError instance.

dburles:
Patch for `defaultSkipError`, which should return `true` if the error is an `EnvelopError` instance.


dotansimha:
thanks @dburles , released as `@envelop/sentry@1.5.1`

PR #1102, opened 2021-12-15
feat: Allow userland code to specify type of currentContext #1101

### LINKED ISSUE #1101: Improve ergonomics around context factory typing
Typing context factory is a bit clunky since `currentContext` is hardcoded to `unknown`:

Screenshot 2021-12-15 at 09 32 01

We are writing a server library based on envelop, and we provide a few things in context out of the box to userland services. We have a priori knowledge that we will already have extended context with `req`, `correlationId` and `logger`.

klippx:
## Description

Allow userland code to get better typings from `currentContext` when defining context factory functions.

Fixes #1101 

With this change we will be able to do this:

Screenshot 2021-12-15 at 09 54 11

Which means our server library did a better job at abstracting the underlying envelop machinery. The userland service only need to worry about their types (`Clients` and `FactoryContext`) while also being provided things already available in the context.

(The `MergedContext` type is not important, but is used for `contextType` in codegen)

## Type of change

### DIFF
--- a/.changeset/fresh-gifts-serve.md
+++ b/.changeset/fresh-gifts-serve.md
@@ -1,0 +1,5 @@
+---
+'@envelop/core': patch
+---
+
+Allow userland code to specify type of currentContext in ContextFactoryFn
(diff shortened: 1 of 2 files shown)


n1ru4l (approved):

dotansimha (approved):

dotansimha:
Thanks @klippx ! :)

PR #1104, opened 2021-12-15
feat: hash with SHA256 cross-platform

### LINKED ISSUE #1084: useResponseCache crypto module usage
The response cache plugin currently uses the crypto module for generating hashes which is only available on Node.js. For true cross platform we should investigate alternatives.

n1ru4l:
Closes https://github.com/dotansimha/envelop/issues/1084

### DIFF
--- a/.changeset/chilled-donuts-search.md
+++ b/.changeset/chilled-donuts-search.md
@@ -1,0 +1,8 @@
+---
+'@envelop/response-cache': minor
+---
+
+Add cross-platform support for platforms that do not have the `Node.js` `crypto` module available by using the `WebCrypto` API. This adds support for deno, cloudflare workers and the browser.
+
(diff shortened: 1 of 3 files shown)


.changeset/chilled-donuts-search.md
```diff
@@ -0,0 +1,8 @@
+---
+'@envelop/response-cache': minor
```
dotansimha, line 2: @n1ru4l should be major?

dotansimha (approved):
LGTM. @n1ru4l please check my comment on the changeset

.changeset/chilled-donuts-search.md
n1ru4l: pre `1.x.x` a minor bump is actually considered a major as long as the package is unstable (see https://semver.org/#spec-item-4)

I am sure we will do a few more breaking changes to the response cache, before making it a full v1. We still need a bit more feedback from the community.

dotansimha: Oh, i see, makes sense, thanks!

n1ru4l:
@dotansimha I haven't tested this on deno or cloudflare workers. We definitely need some kind of automated strategy for doing this. However, the code runs just fine on Node.js and the browser.

PR #1109, opened 2021-12-15
fix(federation): do not cache parse if document is already parsed

ardatan:
Related https://github.com/dotansimha/envelop/discussions/1040


dotansimha:
@ardatan released as `@envelop/apollo-federation@0.2.2`

PR #1125, opened 2021-12-18
feat: validation cache perf

n1ru4l:


packages/plugins/validation-cache/src/index.ts
```diff
@@ -23,8 +25,11 @@ export const useValidationCache = (pluginOptions: ValidationCacheOptions = {}):
     onSchemaChange() {
       resultCache.clear();
     },
-    onValidate({ params, setResult }) {
-      const key = print(params.documentAST);
+    onParse({ params, extendContext }) {
+      extendContext({ [rawDocumentSymbol]: params.source.toString() });
+    },
+    onValidate({ params, context, setResult }) {
+      const key: string = context[rawDocumentSymbol] ?? print(params.documentAST);
```
n1ru4l, line 32: I still kept the `?? print` fallback. Thoughts on removing it completely and instead raising an error?

n1ru4l: rawDocumentSymbol wouldn't be within the context in case the document got never parsed with the `parse` function returned from `getEnveloped`.

dotansimha: I think the fallback is fine just to make sure it works if it's not there

dotansimha (approved):

saihaj (approved):

PR #1140, opened 2021-12-23
Fix memory leak

n1ru4l:
The previous implementation did not properly proxy the `.return` and `.throw` calls, which could lead to consumed event sources not being cleaned up properly.

### DIFF
--- a/.changeset/mean-actors-exist.md
+++ b/.changeset/mean-actors-exist.md
@@ -1,0 +1,5 @@
+---
+'@envelop/core': patch
+---
+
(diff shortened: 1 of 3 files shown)


dotansimha (approved):

n1ru4l:
Somehow this does not seem to be an issue according to https://github.com/dotansimha/envelop/pull/1143

n1ru4l:
@ardatan @dotansimha @saihaj I finally figured out the actual problem and added test cases for that specific problem, that should prevent future regressions 🎉

saihaj (approved):

PR #1237, opened 2022-01-28
feat: useImmediateIntrospection

n1ru4l:
Demonstration for avoiding context creation for introspection operations.

### DIFF
--- a/packages/core/src/orchestrator.ts
+++ b/packages/core/src/orchestrator.ts
@@ -5,6 +5,7 @@ import {
   EnvelopContextFnWrapper,
   GetEnvelopedFn,
   OnContextBuildingHook,
+  AfterContextBuildingHook,
   OnEnvelopedHook,
   OnExecuteDoneHook,
   OnExecuteHook,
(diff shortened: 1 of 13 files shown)


saihaj (approved):

packages/core/src/orchestrator.ts
```diff
@@ -257,22 +258,30 @@ export function createEnvelopOrchestrator<PluginsContext extends DefaultContext>
...
           let context = orchestratorCtx ? { ...initialContext, ...orchestratorCtx } : initialContext;
+          let isBreakingContextBuilding = false;
 
           for (const onContext of beforeCallbacks.context) {
             const afterHookResult = await onContext({
               context,
               extendContext: extension => {
                 context = { ...context, ...extension };
               },
+              breakContextBuilding: () => {
```
dotansimha, line 273: Do we need a general `shortCircuit` API from all steps? 🤔

dotansimha (approved):

packages/core/src/orchestrator.ts
n1ru4l: execute has setExecutionResultAndStopExecution but it also has drawbacks :/ https://github.com/dotansimha/envelop/issues/1156

I think in the future we can add them for all phases 🤔

PR #1328, opened 2022-03-16
Added TInputContext/TOutputContext to plugin API

dotansimha:
Related: https://github.com/dotansimha/envelop/issues/439

### DIFF
--- a/packages/plugins/apollo-tracing/src/index.ts
+++ b/packages/plugins/apollo-tracing/src/index.ts
@@ -39,10 +39,10 @@ type TracingContextObject = {
   hrtime: [number, number];
 };
 
-export const useApolloTracing = (): Plugin => {
+export const useApolloTracing = (): Plugin<{}, { [apolloTracingSymbol]: TracingContextObject }> => {
   return {
     onResolverCalled: ({ info, context }) => {
-      const ctx = context[apolloTracingSymbol] as TracingContextObject;
+      const ctx = context[apolloTracingSymbol];
       // Taken from https://github.com/apollographql/apollo-server/blob/main/packages/apollo-tracing/src/index.ts
       const resolverCall: ResolverCall = {
         path: info.path,
--- a/packages/types/src/hooks.ts
+++ b/packages/types/src/hooks.ts
@@ -68,7 +68,7 @@ export type OnPluginInitEventPayload = {
 export type OnPluginInitHook = (options: OnPluginInitEventPayload) => void;
 
 /** onPluginInit */
-export type OnEnvelopedHookEventPayload<ContextType> = {
+export type OnEnvelopedHookEventPayload<TInputContext, TOutputContext> = {
   /**
    * Set the schema that should be used for execution.
    */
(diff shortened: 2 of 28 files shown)


packages/types/src/plugin.ts
```diff
@@ -11,11 +11,14 @@ import {
   DefaultArgs,
 } from './hooks';
 
-export interface Plugin<PluginContext extends Record<string, any> = {}> {
+export interface Plugin<
+  TInputContext extends Record<string, any> = {},
+  TOutputContext extends Record<string, any> = TInputContext
```
klippx, line 16: Does it work?
```suggestion
  TOutputContext extends TInputContext
```

If so it would be nice to say that you need to return something that at least extends what you were given, without removing anything.

packages/plugins/apollo-tracing/src/index.ts
```diff
@@ -39,10 +39,10 @@ type TracingContextObject = {
   hrtime: [number, number];
 };
 
-export const useApolloTracing = (): Plugin => {
+export const useApolloTracing = (): Plugin<{}, { [apolloTracingSymbol]: TracingContextObject }> => {
```
klippx, line 42: Do you think this `{}` might get interpreted that it _needs_ to be empty in userland, so if userland has some plugin before this it will complain that `Record<string, any>` is not assignable to `{}` ?

n1ru4l: `Record<string, any>` is assignable to `{}`, as `{}` matches basically any object. There is no concept of "exact" types in TypeScript.

Correct me if I am wrong, but it seems like you are expecting `{}` to behave like `Record<string, never>`.

See: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAFASzABwK5QFwwDeAvgJR4BuIaAJkgHxHEBQoksARkqhjvjACUApqABONADzRRGAOYwAPjAgBPALYcQAGwA0MMEIpDR9cjCq0GTZq3DQYAMxAgCwsZOlzFy9Zt3wwFUZkEhs4FCcQUmYOCOdooA

packages/types/src/plugin.ts
n1ru4l: Theoretically, the plugin can overwrite a property in the existing context by calling `extendContext({ property: null })`. So extending might be wrong here.

packages/plugins/apollo-tracing/src/index.ts
klippx: I think I am vaguely referring to an eslint rule that people might use in their project. But I think you are right, I am not sure it would be a problem! 

```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.eslint[@typescript-eslint/ban-types](https://github.com/typescript-eslint/typescript-eslint/blob/v4.22.0/packages/eslint-plugin/docs/rules/ban-types.md)
```

packages/types/src/hooks.ts
```diff
@@ -441,11 +451,11 @@ export type OnSubscribeEventPayload<ContextType> = {
 /**
  * Payload with which the onSubscribeResult hook is invoked.
  */
-export type OnSubscribeResultEventPayload<ContextType> = {
+export type OnSubscribeResultEventPayload<TInputContext, TOutputContext> = {
   /**
    * The execution arguments.
    */
-  args: TypedExecutionArgs<ContextType>;
+  args: TypedExecutionArgs<TInputContext & TOutputContext>;
```
n1ru4l, line 458: Are you sure this is correct? TOutputContext should overwrite context values from TInputContext

TInputContext
```
{
  a: 1,
  b: 2.
}
```

TOutputContext
```
{
  a: 3.
  c: 1,
}
```

Expected Result:

```
{
  a: 3,
  b: 2.
  c: 1,
}
```

Currently, it would result in

```
{
  a: 1 | 3,
  b: 2.
  c: 1,
}
```

packages/types/src/plugin.ts
klippx: Maybe it is something you want to allow? Otherwise it might be nice to have type checks prohibiting accidentally ruining the context chain. 

Wrong:
```
return { property: 'foo' } // type error: { property: 'foo' } is not assignable to TInputContext
```

Right:
```
return { ...inputContext, property: 'foo' }
```

saihaj:
Async discussion with @n1ru4l the conclusion we came to was we don't have a good way to resolve this issue with TypeScript. So will have to think more and revisit.

Issue #1346, opened 2022-03-28
Plugin type unexpectedly gives app context instead of initial context


klippx:
**Describe the bug**
When writing plugins and using certain lifecycle methods you may get tricked into thinking context building has been done and will have app context available to you when it wont be.

**To Reproduce**
See PR with failing spec

**Expected behavior**
I only want to have type support for AppContext when it is actually available

**Environment:**

- OS: MacOS
- NodeJS: v14.18.2
- `@envelop/*` versions:
  - `@envelop/core`: `latest`

klippx:
Related? #1327

PR #1347, opened 2022-03-28
Supply a test that has a type issue

klippx:
## Description

Nothing is fixed, this is a failing (type) check highlighting the issue

Related #1346

## Type of change

### DIFF
--- a/packages/core/test/parse.spec.ts
+++ b/packages/core/test/parse.spec.ts
@@ -1,6 +1,7 @@
 import { assertSingleExecutionValue, createSpiedPlugin, createTestkit } from '@envelop/testing';
-import { FieldNode, parse, visit } from 'graphql';
+import { FieldNode, GraphQLError, parse, visit } from 'graphql';
 import { schema, query } from './common';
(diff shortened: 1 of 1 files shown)


packages/core/test/parse.spec.ts
```diff
@@ -46,6 +47,32 @@ describe('parse', () => {
...
+      onContextBuilding({ extendContext }) {
+        extendContext({
+          favoritePet: 'turtle',
+        });
+      },
+    };
+    const useOnParse: Plugin<AppContext> = {
+      onParse: () => {
+        return ({ context }) => {
+          // @ts-expect-error favoritePet is not available during onParse
```
klippx, line 65: This is the "failing spec", as you can see on [L73](https://github.com/dotansimha/envelop/pull/1347/files#diff-63670871751ebf486fd2051bc6eeb7ffa5001577f118331bd008a3a2fe0b801fR73) this is undefined but TS told me it would be there as it sees no problem with doing `context.favoritePet`

PR #1349, opened 2022-03-28
Pass context to registerContextErrorHandler

### LINKED ISSUE #1348: Pass initialContext to registerContextErrorHandler
When handling errors we want to have access to the initial context which contains the req object.

I would like initialContext to be passed to `registerContextErrorHandler`

I can hack it by sharing a mutable instance variable that gets set somewhere else in the lifecycle and then "hope" that it will be there

klippx:
## Description

Pass inital context to registerContextErrorHandler

Fixes #1348

## Type of change

## Checklist:

### DIFF
--- a/packages/core/src/orchestrator.ts
+++ b/packages/core/src/orchestrator.ts
@@ -268,8 +268,10 @@ export function createEnvelopOrchestrator<PluginsContext extends DefaultContext>
     ? initialContext => async orchestratorCtx => {
         const afterCalls: AfterContextBuildingHook<any>[] = [];
 
+        // In order to have access to the "last working" context object we keep this outside of the try block:
+        let context = orchestratorCtx ? { ...initialContext, ...orchestratorCtx } : initialContext;
+
         try {
-          let context = orchestratorCtx ? { ...initialContext, ...orchestratorCtx } : initialContext;
           let isBreakingContextBuilding = false;
 
           for (const onContext of beforeCallbacks.context) {
@@ -306,7 +308,7 @@ export function createEnvelopOrchestrator<PluginsContext extends DefaultContext>
           let error: unknown = err;
           for (const errorCb of contextErrorHandlers) {
             errorCb({
-              initialContext,
+              context,
               error,
               setError: err => {
                 error = err;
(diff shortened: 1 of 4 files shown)


n1ru4l (approved):

packages/core/src/orchestrator.ts
```diff
@@ -306,6 +306,7 @@ export function createEnvelopOrchestrator<PluginsContext extends DefaultContext>
           let error: unknown = err;
           for (const errorCb of contextErrorHandlers) {
             errorCb({
+              initialContext,
```
n1ru4l: I think it might more sense to actually pass `context` here instead of `initialContext`. As you should probably have access to the "last working" context object (before an error got thrown in case you use multiple context extensions).

n1ru4l (commented):
I added one more comment

packages/core/src/orchestrator.ts
klippx: Agreed! I pushed a change.

One more thing, the type support is a bit lack-luster, but it would be quite difficult to type this correctly.

Do you think improving this seems like a bit too much for this Pr?

Screenshot 2022-04-04 at 12 13 19

You will get that contextSoFar is `unknown` although you as author knows it is of type `string`. The other instances of context are typed as a generic `context: Readonly<ContextType>;` which is a bit large change with a potentially big blast radius, which is why I think it is a bigger effort that be improved later? It is technically also a `Partial<ContextType>` but only `Partial` of those parts that are part of context building Fns... 😵

n1ru4l: his should be addressed as part of https://github.com/dotansimha/envelop/issues/1327

dotansimha (approved):

packages/core/test/context.spec.ts
```diff
@@ -111,4 +112,61 @@ describe('contextFactory', () => {
       })
     );
   });
+
+  it('Should yield initial context to context error handlers', async () => {
```
klippx, line 116:
```suggestion
  it('Should yield context to context error handlers', async () => {
```

PR #1353, opened 2022-03-30
ensure that root query only contains introspection fields

Jomik:
## Description

`useImmediateIntrospection` fails when using `getIntrospectionQuery` from `graphql-js`.
This fixes it.

## Type of change

## How Has This Been Tested?

Locally

## Checklist:

### DIFF
--- a/packages/core/src/plugins/use-immediate-introspection.ts
+++ b/packages/core/src/plugins/use-immediate-introspection.ts
@@ -1,17 +1,5 @@
 import type { Plugin } from '@envelop/types';
-import type { ValidationRule } from 'graphql';
-
-const OnNonIntrospectionFieldReachedValidationRule =
-  (onNonIntrospectionField: () => void): ValidationRule =>
-  () => {
-    return {
-      Field(field) {
-        if (!field.name.value.startsWith('__')) {
-          onNonIntrospectionField();
-        }
-      },
-    };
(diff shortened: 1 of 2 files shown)


Jomik:
Apologies, @n1ru4l I had opened the PR in the wrong account.
> @Jomik  Can you please first add a failing test (using getIntrospectionQuery) and then implement the solution using a validation rule instead of wrapping validate (in order to ensure we have as few traversals as possible)?

I thought that `validate` was only run once, for the entire query? At least the schema and my own test (having `console.log`) showed that.
I am unsure how to implement it using a validation rule, as that is run on every node on some kind, and we actually only want to check the `SelectionSet` on `query`.

n1ru4l:
@Jomik Hey!

I created this alternative solution, can you have a look at https://github.com/dotansimha/envelop/pull/1354

> I thought that validate was only run once, for the entire query? At least the schema and my own test (having console.log) showed that.

Yes, that is true, validate only runs once. My concern is that instead of iterating manually through the document, we can do it as part of the validation rule visitor (and reduce the time complexity). All the validation rules are run with [`visitInParallel`](https://github.dev/graphql/graphql-js/blob/faa42e2cb377cbc26c38f67b2f65891a61c22277/src/language/visitor.ts#L307-L313).

PR #1361, opened 2022-04-04
:boom: NEW: add a failing test when useResponseCache & useGraphQlJit are used together

jycouet:
Maybe it makes sense to have a set of tests mixing plugins?
And/or `createTestkit(...)` having default plugins?


saihaj (approved):

saihaj:
@jycouet try rebasing with `main` and should fix K6 issue

jycouet:
@saihaj thx for the update, but I believe that it's more than just a K6 issue.

PR #1371, opened 2022-04-12
[DO NOT MERGE] Simpler LRU-ish cache implementation

ardatan:
What I realized in Yoga benchmarks, existing LRU cache implementations are blocking the event loop for set/clear operations.
With the usage of `queueMicrotask`, we can push those operations at the end of the current event loop after the request is finished.
Why not `setTimeout` or sth like that because those are pushing the task in the next loop (needs more investigation), so they break serverless envs as I experienced.

Unfortunately I couldn't find enough time to come with real numbers and so on. I'll do once I have time.

### DIFF
--- a/packages/core/src/traced-orchestrator.ts
+++ b/packages/core/src/traced-orchestrator.ts
@@ -3,7 +3,7 @@ import { ArbitraryObject, Maybe } from '@envelop/types';
 import { EnvelopOrchestrator } from './orchestrator';
 import { isAsyncIterable } from './utils';
 
-const getTimestamp = typeof performance !== 'undefined' && performance.now ? () => performance.now() : () => Date.now();
+const getTimestamp = globalThis?.performance?.now ? () => performance.now() : () => Date.now();
 
 const measure = () => {
   const start = getTimestamp();
--- a/packages/core/src/traced-orchestrator.ts
+++ b/packages/core/src/traced-orchestrator.ts
@@ -3,8 +3,7 @@ import { ArbitraryObject, Maybe } from '@envelop/types';
 import { EnvelopOrchestrator } from './orchestrator';
 import { isAsyncIterable } from './utils';
 
-const getTimestamp =
-  typeof globalThis !== 'undefined' && globalThis?.performance?.now ? () => performance.now() : () => Date.now();
+const getTimestamp = typeof performance !== 'undefined' && performance.now ? () => performance.now() : () => Date.now();
 
 const measure = () => {
   const start = getTimestamp();
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -283,8 +283,9 @@ export function getInMemoryLRUCache<V>({
     set(key: string, value: V) {
       queueMicrotask(() => {
         if (storage[key] != null) {
-          if (keys.indexOf(key)) {
-            keys.splice(keys.indexOf(key), 1);
+          const keyIndex = keys.indexOf(key);
+          if (keyIndex > -1) {
+            keys.splice(keyIndex, 1);
           }
         }
         keys.push(key);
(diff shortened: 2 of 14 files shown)


packages/core/src/traced-orchestrator.ts
```diff
@@ -3,8 +3,7 @@ import { ArbitraryObject, Maybe } from '@envelop/types';
 import { EnvelopOrchestrator } from './orchestrator';
 import { isAsyncIterable } from './utils';
 
-const getTimestamp =
-  typeof globalThis !== 'undefined' && globalThis?.performance?.now ? () => performance.now() : () => Date.now();
+const getTimestamp = typeof performance !== 'undefined' && performance.now ? () => performance.now() : () => Date.now();
```
n1ru4l:
```suggestion
const getTimestamp = globalThis?.performance?.now ? () => performance.now() : () => Date.now();
```

I dont think we need a typeof check here globalThis should be everywhere

packages/core/src/utils.ts
```diff
@@ -259,3 +259,64 @@ export function errorAsyncIterator<TInput>(
...
+      });
+    },
+    clear() {
+      queueMicrotask(() => {
+        storage = {};
+        keys = [];
+      });
+    },
+  };
+}
```
n1ru4l: I am unsure whether this should be within core 🤔 
IMHO this could be a library outside of envelop as it is generic enough. I see that it is used within multiple envelop plugins, but not directly within the core.

packages/core/src/utils.ts
```diff
@@ -259,3 +259,64 @@ export function errorAsyncIterator<TInput>(
...
+  let storage: Record<string, V> = {};
+  let keys: string[] = [];
+  return {
+    get(key: string): V | undefined {
+      return storage[key];
+    },
+    set(key: string, value: V) {
+      queueMicrotask(() => {
+        if (storage[key] != null) {
+          if (keys.indexOf(key)) {
```
dotansimha: `indexOf` might return `0` which is false... i guess you need to compare to `-1`?

packages/core/src/utils.ts
```diff
@@ -259,3 +259,64 @@ export function errorAsyncIterator<TInput>(
...
+            keys.splice(keys.indexOf(key), 1);
```
dotansimha: maybe store the value of `keys.indexOf(key)` instead of running it again? :D

packages/core/src/utils.ts
dotansimha: Yeah this can live as external lib (can also be a live under this repo i guess, at least for now)

packages/core/src/utils.ts
ardatan: 🤦 My bad. Fixed

packages/core/src/utils.ts
ardatan: Yes 😅  you're right. Done!

packages/core/src/traced-orchestrator.ts
ardatan: Makes sense!

kamilkisiela:
No more TTL, right? TTL makes not a lot of sense for caching `parse` and `validate` anyway.

packages/plugins/apollo-federation/src/index.ts
```diff
@@ -13,10 +14,19 @@ export interface ApolloFederationPluginConfig {
...
+    set: async (key: string, value: any) => lruCache.set(key, value),
+    delete: async (key: string) => lruCache.delete(key),
+  };
+}
+
 export const useApolloFederation = (options: ApolloFederationPluginConfig): Plugin => {
   const {
     gateway,
-    cache = new InMemoryLRUCache(),
+    cache = getDefaultCache(),
```
kamilkisiela, line 29: maybe TTL makes sense in case of Apollo Gateway?

n1ru4l:
> No more TTL, right? TTL makes not a lot of sense for caching `parse` and `validate` anyway.

That would mean breaking change (MAJOR release), right?

YassinEldeeb:
Hey, @ardatan and @n1ru4l I'm seeing this PR has already been reviewed and I think that'd cause a breaking change release as @n1ru4l has mentioned. what should the next steps be?

PR #1382, opened 2022-04-30
Fix async schema

ardatan:


packages/core/src/plugins/use-schema.ts
```diff
@@ -9,20 +10,47 @@ export const useSchema = (schema: GraphQLSchema): Plugin => {
...
+      const schema$ = schemaLoader(context);
+      if (isPromise(schema$)) {
+        schemaSet$ = schema$.then(schemaObj => {
+          setSchema(schemaObj);
+        });
+      } else {
+        setSchema(schema$);
+      }
+    },
+    onExecute() {
```
n1ru4l: So the idea here is to defer execution until the schema is set - But that will still let validation fail 🤔

ardatan: That's why I left this PR draft :D because only if you don't do `validate` you can use promises for `schema`.

ardatan:
@saihaj closing this one in favor of your PRs

PR #1393, opened 2022-05-12
fix: loosen default context type

n1ru4l:


saihaj (approved):

YassinEldeeb (approved):

packages/types/src/context-types.ts
```diff
@@ -3,7 +3,7 @@ import { Plugin } from './plugin';
 import { Spread, TuplifyUnion, Unarray } from './utils';
 
 // We are using `interface` instead of `type` in order to allow type augmentation
-export interface DefaultContext extends Record<string | symbol | number, unknown> {}
+export interface DefaultContext {}
```
YassinEldeeb: Yeah, this's much cleaner. an empty interface already allows any keys and values by default

YassinEldeeb: just a random code suggestion to trigger the CI to re-run
```suggestion
export interface DefaultContext { }
```

YassinEldeeb:
oh, that's weird. looks like the empty interface isn't behaving correctly. the CI fails with this error:
```sh
error TS2339: Property 'storeId' does not exist on type 'Readonly<DefaultContext>'.
```
which means it doesn't identify `storeId` as a key of the context object which is weird

PR #1406, opened 2022-06-02
:sparkles: NEW: support sentry v7.0.0

jycouet:
Supporting Sentry v7.0.0

`Sentry.Severity.Debug` was depreciated in v6 and removed in v7.
`SeverityLevel` which is the new type to use in v7 was also available in v6.

Release note: [here](https://github.com/getsentry/sentry-javascript/releases/tag/7.0.0)
`SeverityLevel` info [here](https://github.com/getsentry/sentry-javascript/blob/7.0.0/MIGRATION.md#severity-severitylevel-and-severitylevels)

### DIFF
--- a/.changeset/seven-shrimps-attack.md
+++ b/.changeset/seven-shrimps-attack.md
@@ -1,0 +1,5 @@
+---
+'@envelop/sentry': patch
+---
+
+supporting sentry v7.0.0
--- a/packages/plugins/sentry/package.json
+++ b/packages/plugins/sentry/package.json
@@ -31,15 +31,15 @@
   },
   "dependencies": {},
   "devDependencies": {
-    "@sentry/node": "6.18.2",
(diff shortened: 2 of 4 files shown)


jycouet:
🤔 prom-tracing is making the fail, I guess it's this rule => `http_req_duration: ['p(95)<=40']` and the result was: `p(95)=42.01ms`
Is that a real fail?

jycouet:
Is it possible that sentry v7.0.0 got actually slower? 🤔

I took main locally and run benchmark... and p(95) is between `29ms` and `33ms`.
With this branch, I'm between `40ms` and `43ms`!

In the Sentry dashboard (using `6.19.7`), they are asking (nicely) to upgrade to `7.0.0`

Not sure what to do?

n1ru4l (approved):
thank you @jycouet ❤️

PR #1410, opened 2022-06-05
Update prettier, use `@theguild/prettier-config`, add `Lint Prettier` CI action

dimaMachina:


package.json
```diff
@@ -6,26 +6,13 @@
...
-    "./examples/graphql-helix-auth0",
-    "./examples/lambda-aws",
-    "./examples/azure-functions",
-    "./examples/google-cloud-functions",
-    "./examples/cloudflare-workers",
-    "./examples/with-esm",
-    "./examples/typegraphql",
-    "./examples/nexus",
-    "./examples/express-graphql"
+    "./examples/*"
```
dimaMachina: can be simplified with `"./examples/*"`

package.json
```diff
@@ -6,26 +6,13 @@
...
   ],
   "scripts": {
     "prepare": "husky install",
     "postinstall": "patch-package",
     "lint": "eslint --config .eslintrc.json --ext .ts .",
+    "lint:prettier": "prettier --ignore-path .gitignore --check .",
```
dimaMachina, line 29: I guess `lint:prettier` is a better name than `prettier:check`

README.md
```diff
@@ -37,23 +37,21 @@ yarn add graphql @envelop/core
...
+})
 ```
 
 The result of `envelop` is a function that allows you to get everything you need for the GraphQL execution: `parse`, `validate`, `contextBuilder` and `execute`. Use that to run the client's GraphQL queries. Here's a pseudo-code example of how it should look like:
 
 ```ts
 const httpServer = createServer()
 
-httpServer.on('request', (req, res) => {
+httpServer.on('request', async (req, res) => {
```
dimaMachina, line 54: `async` was missing as in line 71 we use await

packages/plugins/response-cache/README.md
```diff
@@ -230,10 +230,10 @@ const getEnveloped = envelop({
   plugins: [
     // ... other plugins ...
     useResponseCache({
-      shouldCacheResult = myCustomShouldCacheResult,
-    }),
-  ],
-});
+      shouldCacheResult: myCustomShouldCacheResult
```
dimaMachina: a typo I guess

package.json
```diff
@@ -79,17 +67,13 @@
   "resolutions": {
     "@changesets/git": "1.3.1",
     "esbuild": "^0.14.0",
-    "prettier": "2.5.1",
```
dimaMachina, line 82: why to have resolutions for prettier?

package.json
```diff
@@ -79,17 +67,13 @@
...
   },
   "lint-staged": {
     "packages/**/src/**/*.{ts,tsx}": [
       "eslint --fix"
     ],
-    "**/*.{ts,tsx,graphql,yml}": [
-      "prettier --write"
-    ],
-    "**/*.json": [
+    "**/*": [
```
dimaMachina: to prettify all files, even markdown

packages/plugins/response-cache/README.md
```diff
@@ -81,28 +81,28 @@ In order to use the Redis cache, you need to:
...
 const redis = new Redis({
   host: 'my-redis-db.example.com',
   port: '30652',
-  password: '1234567890',
-});
+  password: '1234567890'
+})
 
-const redis = new Redis("rediss://:1234567890@my-redis-db.example.com:30652");
+const redis = new Redis('redis://:1234567890@my-redis-db.example.com:30652')
```
B2o5T, line 96: typo

packages/plugins/response-cache/README.md
```diff
@@ -160,55 +160,55 @@ const getEnveloped = envelop({
...
+})
 ```
 
 ### Customize if result should be cached
 
 You can define a custom function used to check if a query operation execution result should be cached.
 
 ```ts
-type ShouldCacheResultFunction = (params: { result: ExecutionResult }) => Boolean;
+type ShouldCacheResultFunction = (params: { result: ExecutionResult }) => Boolean
```
dimaMachina: i guess it's better
```suggestion
type ShouldCacheResultFunction = (params: { result: ExecutionResult }) => boolean
```

website/docs/core.mdx
```diff
@@ -199,17 +199,17 @@ const getEnveloped = envelop({
 This utility is helpful when you want to enable a plugin only when a certain condition is met.
 
 ```ts
-import { envelop, useMaskedErrors, enableIf } from '@envelop/core';
+import { envelop, useMaskedErrors, enableIf } from '@envelop/core'
 
-const isProd = process.env === 'production';
+const isProd = process.env === 'production'
```
dimaMachina: also typo?
```suggestion
const isProd = process.env.NODE_ENV === 'production'
```

packages/plugins/response-cache/README.md
dimaMachina:
```suggestion
const redis = new Redis('rediss://:1234567890@my-redis-db.example.com:30652')
```

package.json
n1ru4l: No, it cannot as some plugins are not part of the workspace (they use the experimental defer & stream). Making them part of the workspace breaks them.

n1ru4l: please revert this change as we have some examples that are not part of the workspace that break due to dependency hoisting.

PR #1417, opened 2022-06-09
Drop official Node.js 12 support

n1ru4l:
Node.js reached end of life a while ago. While we still technically support it we should still move on.

https://endoflife.date/nodejs

image

image


saihaj (approved):

n1ru4l:
Currently blocked by https://github.com/kamilkisiela/bob/issues/42

dotansimha (approved):

saihaj (approved):

PR #1427, opened 2022-06-24
enhance(response-cache) - Use WeakMap instead of context to store document node strings by default and pass all execution args to `getDocumentString`

ardatan:
Currently the document string is added during parse and extracted later during execution. Actually WeakMap can be used here to remove the need of the context.

But still some users of Envelop (e.g. Mesh SDK) don't use Envelop's scoped `parse` function so it is a bit overhead to put printed document node into the context specifically for response cache plugin in order to take from `getDocumentStringFromContext`. In Mesh for example, we have its own `print` and parse` caching so it can give the cached document string by the document node object.

And even if it is not stored and user doesn't provide its own logic , we should still allow the plugin work as is by printing the document node 

I avoided introducing breaking changes but @n1ru4l if you are fine with that, it'd better to make the following breaking changes;

Rename `getDocumentStringFromContext` to `getDocumentString` and remove `context` paramter in favor of all execution parameters (document, variables, context and so on).

Update:
Ok I did some breaking changes by renaming `getDocumentStringFromContext` to `getDocumentString`, also a returned value is required from `getDocumentString` which can be easily fallbacked to graphql-js's print function.

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -18,7 +18,6 @@ import { createInMemoryCache } from './in-memory-cache.js';
 import { hashSHA256 } from './hashSHA256.js';
 
 const contextSymbol = Symbol('responseCache');
-const rawDocumentStringSymbol = Symbol('rawDocumentString');
 
 type Context = {
   identifier: Map<string, CacheEntityRecord>;
@@ -44,7 +43,7 @@ export type BuildResponseCacheKeyFunction = (params: {
   sessionId?: Maybe<string>;
 }) => Promise<string>;
 
-export type GetDocumentStringFromContextFunction = (params: DefaultContext) => Maybe<string>;
+export type GetDocumentStringFromContextFunction = (context: DefaultContext, executionArgs: ExecutionArgs) => Maybe<string>;
 
 export type ShouldCacheResultFunction = (params: { result: ExecutionResult }) => Boolean;
 
@@ -103,11 +102,10 @@ export type UseResponseCacheParameter<C = any> = {
   buildResponseCacheKey?: BuildResponseCacheKeyFunction;
   /**
    * Function used for reading the document string that is used for building the response cache key from the context object.
-   * By default, the useResponseCache plugin hooks into onParse and writes the original operation string to the context.
+   * By default, the useResponseCache plugin hooks into onParse and caches the original operation string in a WeakMap.
    * If you are hard overriding parse you need to set this function, otherwise responses will not be cached or served from the cache.
    * Defaults to `defaultGetDocumentStringFromContext`
    *
-   * @deprecated No longer being used in the next major version.
    */
   getDocumentStringFromContext?: GetDocumentStringFromContextFunction;
   /**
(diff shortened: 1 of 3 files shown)


n1ru4l:
@ardatan breaking change is fine with me 😇

ardatan:
@n1ru4l Ok I updated the PR with breaking changes. Let me know what you think.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -32,18 +32,16 @@ type Context = {
...
-export type GetDocumentStringFromContextFunction = (params: DefaultContext) => Maybe<string>;
+export type BuildResponseCacheKeyFunction = (
+  params: {
+    /** Raw document string as sent from the client. */
+    documentString: string;
+    /** optional sessionId for make unique cache keys based on the session.  */
+    sessionId?: Maybe<string>;
+  } & ExecutionArgs
+) => Promise<string>;
+
```
n1ru4l: Currently we did not supply the context here, because we wanted this function to be as portable as possible to the edge/proxy that runs before the GraphQL API in order to short circuit.

Since most of the `ExecutionArgs` parameters are optional (`contextValue`, `variableValues`), I think it is fine for those. 

However, `document` might not be present on a "edge/proxy" level as it would apply this logic on the raw GraphQL payload. It's value should be optional.

I don't think passing `fieldResolver`, `typeResolver` or `subscribeFieldResolver` to this function provides any value.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -152,8 +151,16 @@ export const defaultShouldCacheResult: ShouldCacheResultFunction = (params): Boo
   return true;
 };
 
-export const defaultGetDocumentStringFromContext: GetDocumentStringFromContextFunction = context =>
-  context[rawDocumentStringSymbol as any] as any;
+const documentStringByDocument = new WeakMap<DocumentNode, string>();
+
+export function defaultGetDocumentString(executionArgs: ExecutionArgs): string {
+  let documentString = documentStringByDocument.get(executionArgs.document);
+  if (!documentString) {
```
n1ru4l: Nit

```
documentString === undefined
```

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -191,15 +198,10 @@ export function useResponseCache({
...
-    onParse(parseCtx) {
-      return function onParseEnd(ctx) {
-        if (ctx.result && 'kind' in ctx.result) {
-          const source = parseCtx.params.source;
-
-          const rawDocumentString = typeof source === 'string' ? source : source.body;
-          ctx.extendContext({
-            [rawDocumentStringSymbol]: rawDocumentString,
-          });
+    onParse({ params: { source } }) {
```
n1ru4l: Nit: In case a custom `getDocumentString` function is set we can omit/skip this `onParse` hook completly?

n1ru4l (changes requested):
@ardatan I added a few comments!

packages/plugins/response-cache/src/plugin.ts
n1ru4l: I agree with the `defaultGetDocumentString ` API change. However, the `BuildResponseCacheKeyFunction` signature should be reverted to its previous implementation as I don't see a benefit in it.

ardatan: Done!

packages/plugins/response-cache/src/plugin.ts
ardatan: Technically yes. Hmm so only if default document string getter is used, onParse is needed. I can refactor there.

n1ru4l: @ardatan thank you

n1ru4l (approved):
Awesome ✨

PR #1428, opened 2022-06-24
enhance(response-cache): Do not touch the schema and respect execution pipeline

ardatan:
Use `visitResult` from `@graphql-tools/utils` instead of modifying the schema by wrapping the resolvers. So if an executor doesn't deal with the schema directly, execution hooks will still be able to process the result only with the execution results passed from the execution.

We also no longer need to interact with the context object because we handle execution result after the execution once.

Sorry I forgot to open this PR as draft because it is still WIP.

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -235,7 +235,8 @@ export function useResponseCache({
               get(_, typename: string) {
                 let typenameCalled = 0;
                 return new Proxy((val: any) => val, {
-                  // Needed for leaf values
+                  // Needed for leaf values such as scalars, enums etc
+                  // They don't have fields so visitResult expects functions for those
                   apply(_, __, [val]) {
                     return val;
                   },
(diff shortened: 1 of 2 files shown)


ardatan:
Need to figure out why this fails;
https://github.com/n1ru4l/envelop/runs/7049628704?check_suite_focus=true#step:9:271

n1ru4l:
@ardatan Can you have a look at the failing test?

ardatan:
Oh I missed that one. I thought this PR was ready. I'll take a look 👀

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -195,41 +181,106 @@ export function useResponseCache({
...
+                  apply(_, __, [val]) {
+                    return val;
+                  },
+                  get(_, fieldName: string) {
+                    if (documentChanged) {
+                      if (fieldName === '__typename') {
+                        typenameCalled++;
+                      }
+                      if (fieldName === '__leave') {
+                        // When typename is defined in the selection set, it calls it twice.
```
n1ru4l: I think we need some inline documentation on why this is the case, I am very confused.

ardatan: I'll add more details with references from graphql-tools source.

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -1482,7 +1482,7 @@ describe('useResponseCache', () => {
     expect(spy).toHaveBeenCalledTimes(1);
   });
 
-  it('cache is purged upon mutation even when error is included in the mutation execution result', async () => {
+  it.skip('cache is purged upon mutation even when error is included in the mutation execution result', async () => {
```
n1ru4l, line 1485: Note that we can no longer do this as we only operate on the complete execution result compared to a resolver level before.

kamilkisiela (approved):

n1ru4l (approved):

PR #1431, opened 2022-06-27
[sentry plugin] Adds an error before using the rootSpan if it's not been set up

orta:
## Description

:wave: It is possible to set up Sentry so that `rootSpan = Sentry.startTransaction` returns null. This is always a user-side sentry setup issue, but because it will crash the plugin I wasn't seeing any logs about what was going on. https://github.com/getsentry/sentry-javascript/issues/4731

Fixes (no issue, didn't seem worth it for a more explicit log)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Too small to warrant, but I did run this code in my app's node_modules first

## Checklist:

- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -191,6 +191,14 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
           op,
           tags,
         });
+
+        if (!rootSpan) {
+          const error = [
+            `Could not create the root Sentry transaction for the GraphQL operation "${transactionName}".`,
+            `It's very likely that this is because you have not included the Sentry tracing SDK in your app's runtime before handling the request.`,
+          ];
+          throw new Error(error.join('\n'));
+        }
       } else {
         const scope = Sentry.getCurrentHub().getScope();
         const parentSpan = scope?.getSpan();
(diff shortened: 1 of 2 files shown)


packages/plugins/sentry/src/index.ts
```diff
@@ -191,6 +191,14 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
           op,
           tags,
         });
+
+        if (!rootSpan) {
+          const error = [
+            `Could not create the root Sentry transaction for the GraphQL operation "${transactionName}".`,
+            `It's very likely that this is because you have not included the Sentry tracing SDK in your app's runtime before handling the request.`,
```
n1ru4l, line 198: @orta Do you think it would be nice to also include a link here that points to the corresponding sentry documentation?

n1ru4l (approved):

packages/plugins/sentry/src/index.ts
n1ru4l: We can do this in a follow-up PR.

PR #1438, opened 2022-07-04
fixes for `import/no-default-export`

dimaMachina:


examples/azure-functions/index.ts
```diff
@@ -20,7 +20,7 @@ const getEnveloped = envelop({
   plugins: [useSchema(schema), useLogger(), useTiming()],
 });
 
-const httpTrigger: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
+export const index: AzureFunction = async (context: Context, req: HttpRequest): Promise<void> => {
```
dimaMachina, line 23: according to docs https://www.npmjs.com/package/@azure/functions

PR #1495, opened 2022-08-22
refactor: isIntrospectionOperation function to not use visit

saihaj:
we don't need to use `visit`


n1ru4l:
does this work with the following query? `{ ... on Query { __schema { __typename }}`?

saihaj:
nvm this is going to be very painful to recursively visit

Issue #1460, opened 2022-07-29
Remove `enableIf`


n1ru4l:
The TypeScript experience if this helper is not good, instead doing `!!condition && usePlugin(condition.props)` often seems better.

PR #1504, opened 2022-09-01
feat: remove `enableIf` utility

saihaj:
closes https://github.com/n1ru4l/envelop/issues/1460


ardatan (approved):

packages/testing/test/test.spec.ts
```diff
@@ -98,7 +97,7 @@ describe('Test the testkit', () => {
       onValidate: jest.fn().mockReturnValue(undefined),
     };
 
-    const testkit = createTestkit([plugin1, enableIf(false, plugin2)], createSchema());
+    const testkit = createTestkit([plugin1, false && plugin2], createSchema());
```
saihaj, line 100: We need to make our types more powerful so we can allow this. Not sure how to do it nicely though cc @n1ru4l @ardatan @dotansimha

n1ru4l (approved):
Let's also add a changeset on why we removed the helper and what people should do instead!

PR #1506, opened 2022-09-02
feat: remove async schema plugin and rename lazy loaded schema plugin

### LINKED ISSUE #1483: Remove useLazyLoadedSchema
This was a mistake from the beginning as we cannot async set the schema (validate and parse) are sync functions.

saihaj:
closes #1483

### DIFF
--- a/packages/core/src/plugins/use-schema.ts
+++ b/packages/core/src/plugins/use-schema.ts
@@ -8,7 +8,7 @@ export const useSchema = (schema: any): Plugin => {
   };
 };
 
-export const useLazyLoadedSchema = (schemaLoader: (context: Maybe<DefaultContext>) => any): Plugin => {
+export const useSchemaByContext = (schemaLoader: (context: Maybe<DefaultContext>) => any): Plugin => {
   return {
     onEnveloped({ setSchema, context }) {
       setSchema(schemaLoader(context));
--- a/packages/core/src/plugins/use-schema.ts
+++ b/packages/core/src/plugins/use-schema.ts
@@ -15,13 +15,3 @@ export const useLazyLoadedSchema = (schemaLoader: (context: Maybe<DefaultContext
     },
   };
 };
-
-export const useAsyncSchema = (schemaPromise: Promise<any>): Plugin => {
-  return {
-    onPluginInit({ setSchema }) {
-      schemaPromise.then(schemaObj => {
-        setSchema(schemaObj);
-      });
-    },
-  };
-};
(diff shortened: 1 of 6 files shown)


website/docs/guides/securing-your-graphql-api.mdx
```diff
@@ -77,25 +77,6 @@ Most of the time this logic should be applied within your business logic that is
 
 Right now envelop ships with two plugins that allow applying authorization before the execution phase.
 
-### Schema based on context
-
-With the `useLazyLoadedSchema` plugin it is possible to dynamically select a schema for execution based on the context object. This is handy if you have a public schema (e.g. for third-party API consumers) and a private schema (for in-house API consumers).
-
-Libraries such as [`graphql-public-schema-filter`](https://github.com/n1ru4l/graphql-public-schema-filter) can be used for generating a schema with only access to a sub part of the original schema using either SDL directives or schema field extensions.
```
n1ru4l, line 84: I think we should still have documentation on how to use a conditional schema

n1ru4l (approved):

packages/core/src/plugins/use-schema.ts
```diff
@@ -7,21 +7,3 @@ export const useSchema = (schema: any): Plugin => {
     },
   };
 };
-
-export const useLazyLoadedSchema = (schemaLoader: (context: Maybe<DefaultContext>) => any): Plugin => {
-  return {
-    onEnveloped({ setSchema, context }) {
-      setSchema(schemaLoader(context));
```
ardatan, line 14: This is sync why do we remove this?

ardatan (changes requested):

packages/core/src/plugins/use-schema.ts
n1ru4l: I have no idea what this even does and the type defs just scream delete to me :D `context: Maybe<DefaultContext>) => any` :D

ardatan: It allows you to pass different schemas on each request which totally makes sense.
```ts
useLazyLoadedSchema(context => {
   if (context.request.headers.get('x-my-token') === 'x-value') {
      return secretschema;
   } else {
      return publicschema;
   }
})
```

That return type `any` is supposed to be `GraphQLSchema` but right now we cannot use types because of "engine agnostic" approach.

n1ru4l: oh `LazyLoaded` indicated for me that stuff should be a promise, maybe we should rename it? `useConditionalSchema`?

ardatan: `useSchemaByContext` :D to make it more obvious

ardatan (approved):

Issue #1482, opened 2022-08-16
Remove `handleValidationErrors` and `handleParseErrors` options from useMaskedErrors


n1ru4l:
`useMaskedErrors` should be scoped to the execution and context creation.

Masking validation errors ONLY makes sense when introspection is disabled.

**ONLY** masking validation errors OR **ONLY** disabling introspection errors does not make sense, as both can be abused for reverse-engineering the GraphQL schema (see https://github.com/nikitastupin/clairvoyance for reverse-engineering the schema based on validation error suggestions).

Thus the best solution here is to actually create a separate plugin that does both, disabling introspection and hiding validation errors.

In terms of masking/hiding parse errors, I am not sure whether this is really needed.

Issue #1343, opened 2022-03-24
[useMaskedErrors] replace EnvelopError with GraphQLError


n1ru4l:
**Is your feature request related to a problem? Please describe.**

Right now we use `EnvelopError` which extends `GraphQLError` for identifying whether an error should be masked or not.

https://github.com/dotansimha/envelop/blob/7e90676ce7b90a1880d5b06e18b6694819aed1fa/packages/core/src/plugins/use-masked-errors.ts#L17

We choose to use a custom `EnvelopError` because the `GraphQLError` constructor API was cumbersome. We changed this within https://github.com/graphql/graphql-js/pull/3454

Instead of 

```ts
throw new GraphQLError("Oops.", undefined, undefined, undefined, undefined, { code: "SERVER_GOES_BRRT" })
```

We can now instantiate a `GraphQLError` with extensions via the new object constructor overload:

```ts
throw new GraphQLError("Oops.", {
  extensions: { code: "SERVER_GOES_BRRT" },
})
```

**Describe the solution you'd like**

Replace `EnvelopError` with `GraphQLError`. This might be a breaking change though we could potentially work around it by just alias exporting `GraphQLError` as `EnvelopError` 🤔 .

```ts
import { GraphQLError } from "graphql";

export const EnvelopError = GraphQLError
```

However, since the constructor changes that would still be a breaking change.

saihaj:
in engine agnostic envelop https://github.com/n1ru4l/envelop/pull/1487 we dropped `useMaskedError` plugin from `core` since that required relying on `GraphQLError` class. Do we want to create a new plug-in for this?

RIP21:
BTW docs are still outdated and say EnvelopError :)

PR #1510, opened 2022-09-05
feat: remove handler for validation and parse errors

saihaj:
Closes https://github.com/n1ru4l/envelop/issues/1343
Closes https://github.com/n1ru4l/envelop/issues/1482

### DIFF
--- a/packages/core/src/plugins/use-masked-errors.ts
+++ b/packages/core/src/plugins/use-masked-errors.ts
@@ -3,80 +3,67 @@ import { handleStreamOrSingleExecutionResult } from '../utils.js';
 
 export const DEFAULT_ERROR_MESSAGE = 'Unexpected error.';
 
-export type FormatErrorHandler = (error: unknown, message: string, isDev: boolean) => Error;
+export type MaskErrorFn = (error: unknown, message: string) => Error;
+
+export type SerializableGraphQLErrorLike = Error & {
+  name: 'GraphQLError';
+  toJSON(): { message: string };
+};
 
 export function isGraphQLError(error: unknown): error is Error & { originalError?: Error } {
   return error instanceof Error && error.name === 'GraphQLError';
 }
 
-function createSerializableError(message: string, originalError?: Error) {
+export function createSerializableGraphQLError(message: string): SerializableGraphQLErrorLike {
   const error = new Error(message);
   error.name = 'GraphQLError';
   Object.defineProperty(error, 'toJSON', {
     value() {
-      if (originalError) {
-        return {
-          message: error.message,
-          extensions: {
-            originalError: {
-              name: originalError.name,
-              message: originalError.message,
-              stack: originalError.stack,
-            },
-          },
-        };
-      }
       return {
         message: error.message,
       };
     },
   });
-  return error;
+  return error as SerializableGraphQLErrorLike;
 }
 
-export const formatError: FormatErrorHandler = (err, message, isDev) => {
+export const defaultMaskErrorFn: MaskErrorFn = (err, message) => {
   if (isGraphQLError(err)) {
     if (err?.originalError) {
       if (isGraphQLError(err.originalError)) {
         return err;
-      } else if (isDev) {
-        return createSerializableError(message, err.originalError);
-      } else {
-        return createSerializableError(message);
       }
+      return createSerializableGraphQLError(message);
     }
     return err;
   }
-  return createSerializableError(message);
+  return createSerializableGraphQLError(message);
 };
 
 export type UseMaskedErrorsOpts = {
-  /** The function used for format/identify errors. */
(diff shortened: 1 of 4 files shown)


packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (err?.originalError) {
+      if (err.originalError.name === 'GraphQLError') {
+        return err as Error;
+      }
+      return new Error(message);
+    }
+    return err as Error;
+  }
+  return new Error(message);
+};
```
n1ru4l: This misses the `originalError` extensions for dev: - https://github.com/n1ru4l/envelop/blob/cea2df214e4d598f1206b1e066bf285960011378/packages/core/src/plugins/use-masked-errors.ts#L30-L37
- https://www.the-guild.dev/graphql/yoga-server/docs/features/error-masking#receive-original-error-in-development

I would prefer to keep this behavior as it can be quite handy.

saihaj: Regular error object doesn’t really have any format for additional metadata

n1ru4l: I think this is highly confusing if the error is suddenly an Error instance instead of an "GraphQLError", that means it is also impossible to add extensions to an masked error? That is a weird limitation as the `formatError` function should imho always return a GraphQL conform JSON serializable error.

n1ru4l:
```bash
node
> JSON.stringify(new Error("lol"))
'{}'
```

this would break `JSON.stringify()` of an execution result.

n1ru4l: Maybe we cannot get around a custom class...

saihaj: Yeah the other strategy is we pull this out from core and provide a plugin for implementation users are using

packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -3,80 +3,67 @@ import { handleStreamOrSingleExecutionResult } from '../utils.js';
 
 export const DEFAULT_ERROR_MESSAGE = 'Unexpected error.';
 
-export type FormatErrorHandler = (error: unknown, message: string, isDev: boolean) => Error;
+export type MaskErrorFn = (error: unknown, message: string) => Error;
+
+export type SerializableGraphQLErrorLike = Error & {
+  name: 'GraphQLError';
+  toJSON(): { message: string };
```
n1ru4l, line 10: I think this should still have extensions as a property in case someone does a console.log 🤔

ardatan: Technically our masking logic doesn't need extensions so it would be extra.

n1ru4l:
I created https://github.com/n1ru4l/envelop/pull/1514 for re-adding extensions with originalError

n1ru4l (changes requested):
This is missing the originalError extensions. I created a POC https://github.com/n1ru4l/envelop/pull/1514

packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -0,0 +1,87 @@
...
+      return createSerializableGraphQLError(message);
+    }
+    return err;
+  }
+  return createSerializableGraphQLError(message);
+};
+
+export type UseMaskedErrorsOpts = {
+  /** The function used for identify and mask errors. */
+  maskErrorFn?: MaskErrorFn;
```
n1ru4l: I personally don't like the `Fn` appendix, Could we name this just MaskError and maskError?

packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -0,0 +1,87 @@
...
+  (maskErrorFn: MaskErrorFn, message: string) =>
+  ({ result, setResult }: { result: ExecutionResult; setResult: (result: ExecutionResult) => void }) => {
+    if (result.errors != null) {
+      setResult({ ...result, errors: result.errors.map(error => maskErrorFn(error, message)) });
+    }
+  };
+
+export const useMaskedErrors = (opts?: UseMaskedErrorsOpts): Plugin => {
+  const maskErrorFn = opts?.maskErrorFn ?? defaultMaskErrorFn;
+  const message = opts?.errorMessage || DEFAULT_ERROR_MESSAGE;
```
n1ru4l, line 86: Do we need this option, we can instead instruct people to use their own masking function 🤔

saihaj: That is true but in some cases it can just be simple to give error message and not have to write 3 extra lines

n1ru4l (approved):

PR #1514, opened 2022-09-07
feat: original error as extensions

n1ru4l:


ardatan (approved):

saihaj (approved):

packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -6,39 +6,64 @@ export const DEFAULT_ERROR_MESSAGE = 'Unexpected error.';
 export type MaskErrorFn = (error: unknown, message: string) => Error;
 
 export type SerializableGraphQLErrorLike = Error & {
-  name: 'GraphQLError';
-  toJSON(): { message: string };
+  toJSON?(): { message: string };
```
ardatan, line 9: Wait I just realized this. Why we remove `name` here? GraphQLError should have `GraphQLError` name as in the following `isGraphQLError` check

packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -6,39 +6,64 @@ export const DEFAULT_ERROR_MESSAGE = 'Unexpected error.';
...
+  extensions?: Record<string, unknown>;
 };
 
 export function isGraphQLError(error: unknown): error is Error & { originalError?: Error } {
```
ardatan, line 13: ~~the return type should be `error is SerializableGraphQLErrorLike`~~

PR #1490, opened 2022-08-18
feat: `useApolloInlineTrace` plugin for Apollo's federated tracing

enisdenjo:
Closes #1489

- [x] Manipulate the final result `extensions` field on parsing and validation errors ([see failing tests](https://github.com/enisdenjo/envelop/blob/36be9caf5063a69bbec6b57f270f255c9df37541/packages/plugins/apollo-inline-trace/test/use-apollo-inline-trace.spec.ts#L206-L240))

PR #1515, opened 2022-09-09
feat: `perform` function that does parsing, validation, context assembly and execution/subscription

### LINKED ISSUE #1493: A single function that takes GraphQL parameters and does parsing, validation, context assembly and execution/subscription
Envelop is currently split into parts and requires the user to integrate them in order every time they want to use envelop and have a valid GraphQL execution chain.

In essence, the redundancy is:
```js
const { parse, validate, contextFactory, execute, schema } = getEnveloped({ req })

const { query, variables } = JSON.parse(payload)

try {
  const document = parse(query)
} catch (err) {
  return { errors: [err] }
}

const validationErrors = validate(schema, document)
if (validationErrors.length > 0) {
  return { errors: validationErrors }
}

const context = await contextFactory(req)
const result = await execute({
  document,
  schema,
  variableValues: variables,
  contextValue: context
})

return result
```

In order to properly process a GraphQL request, you'll always have to parse, validate and execute/subscribe; with this in mind, I suggest a single, unified, `perform` function that does all of the above:

```js
const { perform } = getEnveloped({ req });

const { operationName, query, variables, extensions } = JSON.parse(payload);

const result = await perform({ operationName, query, variables, extensions });

return JSON.stringify(result);
```

This approach, not only simplifies the usage of envelop, but additionally allows manipulation of the final result through an accompanying `onPerform` hook (including parsing and validation errors) to inject implementor specific requirements - like the `extensions` field for Apollo's FTV1 tracing.

Supersedes #1491 
Necessary for #1490

enisdenjo:
Closes #1493

Necessary for #1490

### `perform` function usage
```ts
const { perform } = getEnveloped({ initial: 'ctx' });

const { operationName, query, variables } = JSON.parse(payload);

// Will never throw GraphQL errors
const result = await perform({ operationName, query, variables }, { extension: 'ctx' });

// Result will contain errors for parsing/validation issues or the execution/subscription result(s)
return JSON.stringify(result);
```

### `onPerform` and `onPerformDone` plugin usage
```ts
import { Plugin } from '@envelop/core';

function usePerform(): Plugin {
  return {
    onPerform({ context, extendContext, params, setParams, setResult }) {
      // before performing

      // context: assembled context
      // extendContext: extend assembled context
      // params: GraphQL parameters passed to perform
      // setParams: replace GraphQL parameters before performing
      // setResult: sets an early result for immediate response

      return {
        onPerformDone({ result, setResult }) {
          // after performing

          // result: perform result, contains errors or result(s)
          // setResult: replace the result
        },
      };
    },
  };
}
```

### TODO

- [x] Documentation


saihaj (approved):

ardatan (approved):

n1ru4l (changes requested):
How do I use the `onPerformDone` with Apollo Server?

- The examples need to be updated.
- Should perform be used within the testkit/replace the testkit?

enisdenjo:
> How do I use the `onPerformDone` with Apollo Server?

I've changed the Apollo Server example with https://github.com/n1ru4l/envelop/pull/1515/commits/827fb179a91e8ec89b7c8460c09bc650eb112b72.

> The examples need to be updated.

Not all examples support the `perform` function. Most of them require the isolated GraphQL functions.

I updated those that have support, `simple-http` and `apollo-server`.

> Should perform be used within the testkit/replace the testkit?

I thought about this too, but didn't want to overdo this PR - thought about making a separate one. But, if you want, I can pack it into here.

n1ru4l:
> I thought about this too, but didn't want to overdo this PR - thought about making a separate one. But, if you want, I can pack it here.

I think this should be in here, as `onPerformDone` seems to now be mandatory for some plugins, otherwise, we need to mark what phases each plugin hooks into explicitly in order to let people know what to expect.

I think teaching people that they have to call this (additional?) function with the result is the easiest way.

examples/apollo-server/index.ts
```diff
@@ -24,13 +24,11 @@ const getEnveloped = envelop({
...
-    const { schema, execute, contextFactory } = getEnveloped({ req: requestContext.request.http });
+    const { perform } = getEnveloped({ req: requestContext.request.http });
 
-    return execute({
-      schema,
-      document: requestContext.document,
-      contextValue: await contextFactory(),
-      variableValues: requestContext.request.variables,
+    return perform({
+      query: requestContext.document,
```
ardatan: Not sure about this one because Apollo already runs `validate`, and this will make it called twice, no?
Also are we ok to accept `DocumentNode` to `perform` because this bypasses our parse?

enisdenjo: Ah ok, well, in that case - we should use `execute` instead.

enisdenjo: Or maybe disable the validation in Apollo? 🤔

enisdenjo: We cannot disable parse and validate in Apollo's Server; so, using `getEnveloped.execute` is the way to go here.

n1ru4l (changes requested):
There is still no way for applying the `onPerformDone` logic when not using the `perform` function. Which makes it impossible for users that do not use `perform` to leverage plugins that use the `onPerformDone` hook 🤔 .

I think we should provide such an path as well.

enisdenjo:
I added the `perform` function to the testkit and refactored all compatible tests to use it. Things to note:
- Replaced all usages of `execute` with `perform`
- `mockPhase` cannot be used with `perform`; **however**, no tests use it - does the testkit even need it? Other users?
- Since perform catches errors and puts them inside the result, the Auth0 plugin test changed a bit: [see here](https://github.com/n1ru4l/envelop/pull/1515/commits/288fd29812c7974658248782aa28ccd8512b16af#diff-73068ab1e7ac5e7c8362ea12f4f83852d6c2a5a01a001de57b3b4c2fed9ba907L332-R344).

> There is still no way for applying the onPerformDone logic when not using the perform function. Which makes it impossible for users that do not use perform to leverage plugins that use the onPerformDone hook 🤔 .

@n1ru4l that is the thing, if you don't use `perform` you cannot use `onPerform` (`onPerformDone` neither because it's returned from `onPerform`). Exactly because of this, I put `onPerformDone` in `onPerform` instead of it being a top level hook.

Perform is the only function in envelop that wraps parsing and validation errors - you _must_ use it to leverage plugins using these hooks. The problem is described in #1493.

Do you have some alternatives maybe?

EmrysMyrddin:
If I understand well what n1ru4l says, it's not a matter of using or not `perform`.
It's more about plugin developers not being able to use this hook since it can't know in advance if there users will use `perform` or the original functions. This makes the hook difficult to use and probably not very usefull for most plugins that aims to be reused and distributed as a package.

enisdenjo:
> It's more about plugin developers not being able to use this hook since it can't know in advance if there users will use perform or the original functions. This makes the hook difficult to use and probably not very usefull for most plugins that aims to be reused and distributed as a package.

Yes, I understand this too. But I don't see a lot of feasible solutions, maybe one of them would be: log a warning (throw error? too harsh?) if a plugin uses the perform hooks but the invoker didn't use perform. 🤔

enisdenjo:
What if we augment each gql function so that it calls `onPerformDone` so that plugins still get some result manipulation capabilities?

- `parse` - catch throws, format as exec result and run through hooks. If the invoker is parse from getEnvloped, take first error from the exec result and bubble it instead
- `validate` - caught or returned errors are formatted as exec result and ran through hooks. If invoker is from getEnveloped, return just the errors from the exec result

This way, users can notice when their results are incomplete with respectful plugins and read about how to fix it - the plugins should state this.

However, if the plugin doesn't expect changing full exec results for parsing and validating - nobody will notice.

Still there is this question: how to invoke `onPerform`? Or alternatively inline `onPerformDone`?

PR #1520, opened 2022-09-12
docs: migration guide

saihaj:


website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+### 6. Drop `useTiming` plugin
+
+### 7. Drop support for Node.js v12
+
+Node.js v12 is no longer supported by the Node.js team. https://github.com/nodejs/Release/#end-of-life-releases
+
+### 8. Remove `onResolverCalled`
+
+We were wrapping the `GraphQLSchema` object but with the new version we no longer want to depend on a specific implmentation. So we decided to drop this and instead provide a new plugin that will let you hook into this phase.
```
enisdenjo:
```suggestion
We were wrapping the `GraphQLSchema` object but with the new version we no longer want to depend on a specific implementation. So we decided to drop this and instead provide a new plugin that will let you hook into this phase.
```

website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+function useResolve(): Plugin {
+  return {
+-   onResolverCalled: onResolverCalled,
++   onPluginInit: ({ addPlugin }) => {
++     addPlugin(useOnResolve(onResolverCalled));
++   },
+  };
+}
+
+const getEnveloped = envelop({
```
enisdenjo: Maybe also pass in the `parse, validate, execute, subscribe` from graphql? Just to keep it in line with the rest of the document.

enisdenjo (changes requested):

website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+import { envelop } from '@envelop/core';
++ import { parse, validate, execute, subscribe } from 'graphql';
+
+- const getEnveloped = envelop([ ... ])
++ const getEnveloped = envelop({ parse, validate, execute, subscribe, plugins: [ ... ] })
+```
+
+### 2. Remove `useAsyncSchema` plugin
+
+This was a mistake from beginning as we cannot asynchronously `validate` and `parse` since with [`graphql`](https://github.com/graphql/graphql-js) these functions are synchronous in nature.
```
n1ru4l, line 63: misses documentation path for correctly using an async loaded schema

website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+   ]
+ })
+```
+
+### 5. Update options for `useMaskedErrors` plugin
+
+- We _removed_ `handleValidationErrors` and `handleParseErrors` options since ONLY masking validation errors OR ONLY disabling introspection errors does not make sense, as both can be abused for reverse-engineering the GraphQL schema (see https://github.com/nikitastupin/clairvoyance for reverse-engineering the schema based on validation error suggestions). Instead you should use `useErrorHandler` plugin where you can access each phase and decide what to do with the error.
+- Renamed `formatError` to `maskError`
+
+### 6. Drop `useTiming` plugin
```
n1ru4l: What do we tell people that still want to use it?

website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+-    enableIf(isProd, useMaskedErrors())
++    isProd && useMaskedErrors()
+   ]
+ })
+```
+
+### 5. Update options for `useMaskedErrors` plugin
+
+- We _removed_ `handleValidationErrors` and `handleParseErrors` options since ONLY masking validation errors OR ONLY disabling introspection errors does not make sense, as both can be abused for reverse-engineering the GraphQL schema (see https://github.com/nikitastupin/clairvoyance for reverse-engineering the schema based on validation error suggestions). Instead you should use `useErrorHandler` plugin where you can access each phase and decide what to do with the error.
+- Renamed `formatError` to `maskError`
```
n1ru4l, line 114: Misses an code example from migrating from v2 to v3

website/docs/guides/migrating-from-v2-to-v3.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+const getEnveloped = envelop({
+  plugins: [
+    // ... other plugins ...
+    useResolve(),
+  ],
+});
+```
+
+PR: https://github.com/n1ru4l/envelop/pull/1500
```
notrab: Make move this up top so the first line is [new major version](https://github.com/n1ru4l/envelop/pull/1500)

n1ru4l:
Some feedback:
- No mention of removing EnvelopError
- Misses how people should migrate off from `instanceof` EnvelopError/GraphQLError checks
- misses why we removed orchestrator tracing

website/docs/guides/migrating-from-v2-to-v3.mdx
saihaj: Confused? We removed async loaded schema cause it was never “async” in nature

website/docs/guides/migrating-from-v2-to-v3.mdx
saihaj: We don’t have alternative for this yet and honestly don’t think we should make this for real world prod usage most people would use sentry, Prometheus or Hive to get the resolver usage timings

website/docs/guides/migrating-from-v2-to-v3.mdx
saihaj: do you have something specific?

enisdenjo (approved):

website/docs/guides/migrating-from-v2-to-v3.mdx
n1ru4l:
```
const schema = await buildSchem()

const envelop = getEnveloped({plugins: useSchema(schema)})
```

I meant sth like this 😅

PR #1487, opened 2022-08-17
V3

saihaj:
- [x] Change `envelop` function structure to require passing all functions
- [x] Use TS structural typings for functions signatures
- [X] Remove  `graphql` import from `core` and drop as `peerDependencies`
- [x] Ensure we have eslint rule to prevent imports from `graphql` 
- [x] Run tests with the fork
- [x] Remove `graphql` imports in plugins where possible - https://www.notion.so/theguildoss/Envelop-Plugins-Engine-Agnostic-5e055afb91284365901bac973d353f95
- [x] Remove GraphQL matrix from CI pipeline
- [x] Major breaking change changeset 
- [x] Update docs
- [X] versioned docs with `v3` default

Goal is to have this kind of migration path:

#### Before

```ts
const getEnveloped = envelop([ ... ])
```

#### After

```ts
import { parse, validate, execute, subscribe } from 'graphql';

const getEnveloped = envelop({ parse, validate, execute, subscribe, plugins: [ ... ] })
```

PR #1525, opened 2022-09-16
fix replace resolver persistence

### LINKED ISSUE #1524: ReplaceResolverFn is replacing resolve is mutating field definition
### Issue workflow progress

_Progress of the issue based on the [Contributor Workflow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow)_

- [ ] 1. The issue provides a [minimal reproduction](https://en.wikipedia.org/wiki/Minimal_reproducible_example) available on [Stackblitz](https://stackblitz.com/fork/node).
- [x] 2. A failing test has been provided
- [x] 3. A local solution has been provided
- [x] 4. A pull request is pending review

---

**Describe the bug**

When implementing a plugin, it is possible to replace the resolver function just before it executed, based an context or any other request information.

The problem is that this replacement is defintive and is persistent over following requests.

**To Reproduce**
Here a dummy example of the bug (dummy because of course, here, we have better ways to pass an authenticated context).
In this example, once an authenticated request have been executed, every non authenticated request would also receive the authenticated context.
```js
const getEnvelop = envelop({
	plugins: {
		onResolverCalled({ context, resolverFn, replaceResolverFn }) {
			if (context.isAuthenticated) return
			replaceResolverFn(
				(...args) => resolverFn(...args, authenticatedContext)
			)
		}
	}
}
```

**Expected behavior**

I expect the resolver function to be reset to it's original implementation for each request since this hook is called for each requests

**Additional context**

My example here is very simple, my actual plugin is doing things that cannot be done in an other way. It's goal is to allow field resolve functions to be generator functions, allowing to use a coroutine library (https://github.com/cuillerejs/cuillere).

You can find the current implementation of the envelop integration here : https://github.com/cuillerejs/cuillere/blob/main/packages/envelop/src/envelop.ts

EmrysMyrddin:
## Description

This PR aims to fix the feature allowing to replace the resolver function in place just before it is called.

Since the `onResolverCalled` hook is being called on each resolver calls, it is expected that the `replaceResolverFn` function only applies to this resolver call. This PR implement this behaviour.

Fixes #1524

## Type of change

	- Since this is changing the behaviour of `replaceResolverFn`, it possible that it introduce some unexpected breaking changes if some relied on the persistent replacement.
	- Since this is an undocumented feature, there isn't any documentation update needed for now.

## How Has This Been Tested?

I have added test cases to the test suite.

**Test Environment**:

- OS: macOS 12.5.1
- `@envelop/core`: 98840fda07b1bc1d6d84d18c6da01eebcb096d88
- NodeJS: 16.14.0

## Checklist:

### DIFF
--- a/.changeset/hip-beds-hear.md
+++ b/.changeset/hip-beds-hear.md
@@ -1,0 +1,5 @@
+---
+'@envelop/core': patch
+---
+
+fix replace resolver fn persistence
--- a/packages/core/src/traced-schema.ts
+++ b/packages/core/src/traced-schema.ts
@@ -17,12 +17,13 @@ export function prepareTracedSchema(schema: GraphQLSchema | null | undefined): v
       const fields = Object.values(type.getFields());
 
       for (const field of fields) {
-        let resolverFn: ResolverFn = (field.resolve || defaultFieldResolver) as ResolverFn;
+        const originalResolver: ResolverFn = (field.resolve || defaultFieldResolver) as ResolverFn;
 
         field.resolve = async (root, args, context, info) => {
           if (context && context[resolversHooksSymbol]) {
             const hooks: OnResolverCalledHook[] = context[resolversHooksSymbol];
             const afterCalls: AfterResolverHook[] = [];
+            let resolverFn = originalResolver;
 
             for (const hook of hooks) {
               const afterFn = await hook({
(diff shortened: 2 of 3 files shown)


saihaj:
Thanks for the PR! while working on `v3` we also fixed this issue :) cc  @ardatan https://github.com/n1ru4l/envelop/pull/1487

PR #1528, opened 2022-09-17
give access to execute args in generic auth plugin

### LINKED ISSUE #1527: [@envelop/generic-auth] Give access to execute args in the validateUser function

I wants to create an auth plugin which allow to define custom validation for each field based on arguments.

I want to have access to execution args in the user validation to have access to context and variables.

Here a simple example:
```js
const getEnvelop = envelop({
	plugins: [
		useGenericAuth({
			mode: 'protect-all',
			directiveOrExtensionFieldName: 'auth',
			resolveUserFn: (ctx) => getUser(ctx.authToken)
			validateUser: async ({ user, fieldAuthExtension, args }) => {
				const validate = fieldAuthExtension?.validate
				if(validate) {
					await validate({ user, ctx: args.contextValue, variables: ctx.variableValues })
				}
			}
		})
	]
})

const resolvers = {
	Query: {
		user: {
			resolve: (_, { userId }) => getUser(userId),
			extensions: {
				auth: {
					validate: ({ user, variables }) => {
						if(user.id !== variables.userId) throw new Error("Unauthorized")
					}
				}
			}
		}
	}
}
```

An alternative would be to do this custom checks directly in resolvers. But I think it's clearer to have access authorization uncoupled from data fetching.

EmrysMyrddin:
## Description

This PR aims to give more flexibility to the user validation of the generic-auth plugin byt giving access to the execute args.
This would give access to context and variables of the operation.

Fixes #1527

## Type of change

## How Has This Been Tested?

I have published this PR as `@emrys-myrddin/envelop-generic-auth@4.5.0-fork.1` and tested it on my company's project.
I was able to build a plugin leading to the api I have describe in the issue #1527.

**Test Environment**:

- OS: macOS 12.5.1
- `@envelop/generic-auth`: v4.5.0
- NodeJS: v16.14.0

## Checklist:

  - The documentation is not very exhaustive on available arguments, I'm not sure it is needed to modify it. I've just updated code documentation.
  - Not sure any test are needed here since types already ensure that this works

### DIFF
--- a/packages/plugins/generic-auth/src/index.ts
+++ b/packages/plugins/generic-auth/src/index.ts
@@ -1,6 +1,7 @@
 import { DefaultContext, Maybe, Plugin, PromiseOrValue } from '@envelop/core';
 import {
   DirectiveNode,
+  ExecutionArgs,
   FieldNode,
   getNamedType,
   GraphQLError,
@@ -30,6 +31,8 @@ export type ValidateUserFnParams<UserType> = {
   fieldAuthDirectiveNode: DirectiveNode | undefined;
   /** The extensions used for authentication (If using an extension based flow). */
   fieldAuthExtension: unknown | undefined;
+  /** The args passed to the execution function (including operation context and variables) **/
+  args: ExecutionArgs;
 };
 
 export type ValidateUserFn<UserType> = (params: ValidateUserFnParams<UserType>) => void | UnauthenticatedError;
@@ -176,6 +179,7 @@ export const useGenericAuth = <
                     objectType,
                     fieldAuthDirectiveNode,
                     fieldAuthExtension,
+                    args,
                   });
                   if (error) {
                     context.reportError(error);
--- a/packages/plugins/generic-auth/src/index.ts
+++ b/packages/plugins/generic-auth/src/index.ts
@@ -32,7 +32,7 @@ export type ValidateUserFnParams<UserType> = {
   /** The extensions used for authentication (If using an extension based flow). */
   fieldAuthExtension: unknown | undefined;
   /** The args passed to the execution function (including operation context and variables) **/
-  args: ExecutionArgs;
+  executionArgs: ExecutionArgs;
 };
 
 export type ValidateUserFn<UserType> = (params: ValidateUserFnParams<UserType>) => void | UnauthenticatedError;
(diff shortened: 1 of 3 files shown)


packages/plugins/generic-auth/src/index.ts
```diff
@@ -176,6 +179,7 @@ export const useGenericAuth = <
                     objectType,
                     fieldAuthDirectiveNode,
                     fieldAuthExtension,
+                    args,
```
n1ru4l: Can we give this a better naming like `executionArgs`?

n1ru4l (changes requested):
Only a small renaming thing, please also update the corresponding documentation and add an example to the changeset ☺️

EmrysMyrddin:
Where should I add documentation and example for this ?

n1ru4l:
@EmrysMyrddin This is the plugin documentation: https://www.envelop.dev/plugins/use-generic-auth

EmrysMyrddin:
So you want me to add a section after https://www.envelop.dev/plugins/use-generic-auth#extend-authentication-with-custom-directive-logic with an example usage of the execution arguments ?

n1ru4l:
Yes and please also update the changeset to give an code example as well as mention a possible use-case where this is helpful.

EmrysMyrddin:
I'm not sure to understand what you mean by this sorry. I can add examples to the changeset ? It's not just a short descripion of my changement ?

n1ru4l:
@EmrysMyrddin You can add a code block example to the changeset, like the following: https://github.com/n1ru4l/envelop/blob/main/packages/core/CHANGELOG.md#minor-changes-7

Just the changelog entry `give access to execute args in generic auth plugin` won't show people how they can now use this.

EmrysMyrddin:
Ho ok that's really nice. I have added the changeset doc + updated the documentation (with some little refactoring to make the "custom logic" part more clear).

EmrysMyrddin:
Writing the documentation made me wondering if the custom per field validation couldn't be directly implemented by the plugin ? I suppose validating a field with arguments against the user is pretty common.

n1ru4l (approved):

n1ru4l:
> Writing the documentation made me wondering if the custom per field validation couldn't be directly implemented by the plugin ? I suppose validating a field with arguments against the user is pretty common.

That is something open for discussion and out of scope of this PR!

PR #1541, opened 2022-10-06
fix integrating-with-databses

dimaMachina:


website/src/pages/docs/guides/integrating-with-databases.mdx
```diff
@@ -169,8 +169,6 @@ const databaseClientPlugin = {
     // We can find in the context the kind of request we are executing
     if (contextValue?.operation?.operation === mutation) {
       await Promise.all(clients.map(client => client.query('BEGIN')))
-    } else {
-      await Promise.all(clients.map(client => client.query('BEGIN')))
```
enisdenjo: I've just checked the whole document, looking at the [Enforcing Read-Only Queries](https://www.the-guild.dev/graphql/envelop/docs/guides/integrating-with-databases#enforcing-read-only-queries) you can notice that queries use a `BEGIN READ ONLY` transaction instead. My guess is that's why the else existed in the first place.

```suggestion
    } else {
      await Promise.all(clients.map(client => client.query('BEGIN READ ONLY')))
```

Furthermore, starting a transaction is necessary because `onExecuteDone` releases them (a transaction must exist) - I am quite confident this is how it should look like.

@dotansimha @n1ru4l please correct me if I am wrong.

enisdenjo (changes requested):

website/src/pages/docs/guides/integrating-with-databases.mdx
dotansimha: Yeah, I agree that `BEGIN READ ONLY` is a good fit here!

website/src/pages/docs/guides/integrating-with-databases.mdx
```diff
@@ -169,8 +169,6 @@ const databaseClientPlugin = {
     // We can find in the context the kind of request we are executing
     if (contextValue?.operation?.operation === mutation) {
       await Promise.all(clients.map(client => client.query('BEGIN')))
```
dimaMachina, line 171:
```suggestion
      await Promise.all(clients.map(client => client.query('BEGIN READ ONLY')))
```

website/src/pages/docs/guides/integrating-with-databases.mdx
dimaMachina: @enisdenjo should I keep `else` statement?

enisdenjo: Yes, should be as I suggested. 👍

dotansimha: @B2o5T I think the latest code is not updated? it should be `BEGIN` for mutation and `BEGIN READ ONLY` for query?

dimaMachina: Yes, will fix it tomorrow, until will convert to draft

enisdenjo (approved):

PR #1560, opened 2022-10-14
Plugins with context generic for correct inheritance

enisdenjo:


saihaj (approved):

saihaj (commented):
is there a way we can do some type tests 🤔 to avoid this in future

PR #1570, opened 2022-10-27
Skip GraphQLError by default in Sentry plugin

kamilkisiela:
Throwing a `GraphQLError` in GraphQL Yoga v3 results in error being masked.
The GraphQLError is treated there as an expected error.
The sentry plugin should follow this logic and do not capture GraphQLError.

It's not the case currently: https://github.com/n1ru4l/envelop/actions/runs/3335641522/jobs/5519849055

WDYT?

### DIFF
--- a/packages/plugins/sentry/test/sentry.spec.ts
+++ b/packages/plugins/sentry/test/sentry.spec.ts
@@ -1,0 +1,9 @@
+import { defaultSkipError } from '../src/index.js';
+import { GraphQLError } from 'graphql';
+
+test('defaultSkipError should return true for GraphQLError', () => {
+  // Throwing a GraphQLError in GraphQL Yoga v3 results in error masking.
+  // The GraphQLError is treated as an expected error.
+  // The sentry plugin should follow this logic and do not capture GraphQLError.
+  expect(defaultSkipError(new GraphQLError('test', {}))).toBe(true);
+});
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -86,10 +86,7 @@ export type SentryPluginOptions = {
 };
 
 export function defaultSkipError(error: Error): boolean {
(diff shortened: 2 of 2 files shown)


packages/plugins/sentry/test/sentry.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { defaultSkipError } from '../src/index.js';
+import { GraphQLError } from 'graphql';
+
+test('defaultSkipError should return true for GraphQLError', () => {
+  // Throwing a GraphQLError in GraphQL Yoga v3 results in error masking.
+  // The GraphQLError is treated as an expected error.
+  // The sentry plugin should follow this logic and do not capture GraphQLError.
+  expect(defaultSkipError(new GraphQLError('test', {}))).toBe(true);
```
n1ru4l, line 8: I don't think this behavior is correct as the graphql execution engine will wrap any error within `GraphQLError`.

Thus throwing a `GraphQLError` from within a resolver becomes a `GraphQLError` with an `originalError` property that is also a `GraphQLError`. In general, all errors whose `originalError` property is a `GraphQLError` is an expected errors that should be reported.

Throwing a non `GraphQLError` from within a resolver would result in a `GraphQLError` that has an `originalError` property of a non `GraphQLError` instance - which would not be reported.

ardatan: This should be used instead then;
https://github.com/n1ru4l/envelop/blob/main/packages/core/src/plugins/use-masked-errors.ts#L18

n1ru4l: it is more than this :D

ardatan: Why?

ardatan: Why?

PR #1571, opened 2022-10-27
enhance(core): check errors recursively if they are originally GraphQLError

ardatan:


packages/core/src/plugins/use-masked-errors.ts
```diff
@@ -15,6 +15,16 @@ export function isGraphQLError(error: unknown): error is Error & { originalError
   return error instanceof Error && error.name === 'GraphQLError';
 }
 
+export function isOriginalGraphQLError(error: unknown): error is Error & { originalError?: Error } {
+  if (isGraphQLError(error)) {
+    if (error.originalError != null) {
+      return isGraphQLError(error.originalError);
```
saihaj: Should we use `isOriginalGraphQLError` then we make it truly recursive?

n1ru4l (approved):

packages/core/src/plugins/use-masked-errors.ts
ardatan: My bad

PR #1582, opened 2022-11-17
test: global token of a module should be accessible by itself (singleton)

kamilkisiela:
Urigo/graphql-modules#2278

The issue: missing `design:paramtypes`

### DIFF
--- a/packages/plugins/graphql-modules/test/use-graphql-modules.spec.ts
+++ b/packages/plugins/graphql-modules/test/use-graphql-modules.spec.ts
@@ -1,6 +1,6 @@
 import 'reflect-metadata';
 import { parse } from 'graphql';
-import { Application, createApplication, createModule, Injectable, Scope } from 'graphql-modules';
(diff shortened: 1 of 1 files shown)


saihaj (dismissed):

n1ru4l:
@saihaj bruh you approved a failing test? :D

saihaj:
> @saihaj bruh you approved a failing test? :D

oops I just saw benchmark which fails time to time didn't realize the actual test also failed

n1ru4l:
@kamilkisiela what are the next steps here?

YassinEldeeb:
Hey, @kamilkisiela that looks great. but from inspecting the tests I'm seeing that the test `use-graphql-modules` is failing because we expected a value to be returned when it returned `null` and the following error:

```sh
{
...,
      errors: [
        TypeError: Cannot read properties of undefined (reading 'lorem')
        ...
      ]
}
```

PR #1598, opened 2022-12-16
fix: include validation rule names in cache key

n1ru4l:
Include the validation rule names within the operation cache key.

This prevents skipping conditional validation rules in other plugins.
Please make sure your validation rules always have a unique `name` property.

**Note:** I would also prefer to hash the validation names... but we currently have no way of doing cross-platform hashing in a sync and performant way. 😅 We could use something like https://github.com/emn178/js-sha1/blob/master/src/sha1.js instead for now.

____

**Note:** In a follow-up PR we should also create a hash key per GraphQLSchema instead of resetting the whole cache in case of a schema change.

### DIFF
--- a/packages/plugins/validation-cache/src/index.ts
+++ b/packages/plugins/validation-cache/src/index.ts
@@ -50,26 +50,30 @@ export const useValidationCache = (pluginOptions: ValidationCacheOptions = {}): 
     onParse({ params, extendContext }) {
       extendContext({ [rawDocumentSymbol]: params.source.toString() });
     },
-    onValidate({ params, context, setResult }) {
-      let ruleKey = '';
-      if (Array.isArray(params.rules)) {
-        // Note: We could also order them... but that might be too much
-        for (const rule of params.rules) {
-          ruleKey = ruleKey + rule.name;
+    onValidate({ params, context, setValidationFn, validateFn }) {
+      // We use setValidateFn over accessing params.rules directly, as other plugins in the chain might add more rules.
+      // This would cause an issue if we are constructing the cache key here already.
+      setValidationFn((...args) => {
+        let ruleKey = '';
+        if (Array.isArray(args[2])) {
+          // Note: We could also order them... but that might be too much
+          for (const rule of args[2]) {
+            ruleKey = ruleKey + rule.name;
+          }
         }
-      }
 
-      const key: string = ruleKey + (context[rawDocumentSymbol] ?? print(params.documentAST));
-      const cachedResult = resultCache.get(key);
+        const key: string = ruleKey + (context[rawDocumentSymbol] ?? print(params.documentAST));
+        const cachedResult = resultCache.get(key);
 
-      if (cachedResult !== undefined) {
-        setResult(cachedResult);
-      }
+        if (cachedResult !== undefined) {
+          return cachedResult;
+        }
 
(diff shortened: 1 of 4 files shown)


packages/plugins/validation-cache/src/index.ts
```diff
@@ -50,18 +50,30 @@ export const useValidationCache = (pluginOptions: ValidationCacheOptions = {}):
...
     },
-    onValidate({ params, context, setResult }) {
-      const key: string = context[rawDocumentSymbol] ?? print(params.documentAST);
-      const cachedResult = resultCache.get(key);
+    onValidate({ params, context, setValidationFn, validateFn }) {
+      // We use setValidateFn over accessing params.rules directly, as other plugins in the chain might add more rules.
+      // This would cause an issue if we are constructing the cache key here already.
+      setValidationFn((...args) => {
+        let ruleKey = '';
+        if (Array.isArray(args[2])) {
```
saihaj, line 58: we always have second `arg` here?

packages/plugins/validation-cache/src/index.ts
```diff
@@ -50,18 +50,30 @@ export const useValidationCache = (pluginOptions: ValidationCacheOptions = {}):
...
+          }
+        }
 
-      if (cachedResult !== undefined) {
-        setResult(cachedResult);
-      }
+        const key: string = ruleKey + (context[rawDocumentSymbol] ?? print(params.documentAST));
+        const cachedResult = resultCache.get(key);
+
+        if (cachedResult !== undefined) {
```
saihaj, line 68:
```suggestion
        if (!!cachedResult) {
```

saihaj (approved):

packages/plugins/validation-cache/src/index.ts
n1ru4l: Well, if we don't have access - we cannot create a cache key. 😅
I assume because of the `useEngine` call, these rules always exist:

https://github.com/n1ru4l/envelop/blob/29326750382ff963619b30ec91241146458a89bd/packages/core/src/plugins/use-engine.ts#L27

packages/plugins/validation-cache/src/index.ts
```diff
@@ -50,18 +50,30 @@ export const useValidationCache = (pluginOptions: ValidationCacheOptions = {}):
...
+    onValidate({ params, context, setValidationFn, validateFn }) {
+      // We use setValidateFn over accessing params.rules directly, as other plugins in the chain might add more rules.
+      // This would cause an issue if we are constructing the cache key here already.
+      setValidationFn((...args) => {
+        let ruleKey = '';
+        if (Array.isArray(args[2])) {
+          // Note: We could also order them... but that might be too much
+          for (const rule of args[2]) {
+            ruleKey = ruleKey + rule.name;
+          }
```
n1ru4l, line 62: I would also prefer to hash the validation names... but we currently have no way of doing cross-platform hashing in a sync and performant way. 😅 We could use something like [emn178/js-sha1@master/src/sha1.js](https://github.com/emn178/js-sha1/blob/master/src/sha1.js?rgh-link-date=2022-12-16T19%3A54%3A56Z) instead for now.

Thoughts?

dotansimha (approved):

PR #1602, opened 2022-12-22
feat: respect GraphQL schema in the validation cache key

### LINKED ISSUE #1599: [validation-cache] do not evict cache upon schema change
Right now the cache is evicted when the schema is changed. https://github.com/n1ru4l/envelop/blob/29326750382ff963619b30ec91241146458a89bd/packages/plugins/validation-cache/src/index.ts#L43-L49

Instead, it would make more sense to generate a unique hash for each schema and include it within the cache key. In order to avoid hashing the schema over and over again we can utilize a `WeakMap<GraphQLSchema, string>`.

The problem is the cache key calculation. We can run the introspection query on GraphQLSchema, then use `json-fast-stable-stringify` for converting it into a string. In order to reduce the cache key length a bit it would also make sense to hash the string.

However, we could not use the `Crypto` API (cross-platform) for the hashing as validation must be sync and cannot be async.

n1ru4l:
Closes https://github.com/n1ru4l/envelop/issues/1599

### DIFF
--- a/.changeset/@envelop_validation-cache-1602-dependencies.md
+++ b/.changeset/@envelop_validation-cache-1602-dependencies.md
@@ -1,0 +1,8 @@
+---
+'@envelop/validation-cache': patch
+---
+
+dependencies updates:
+
+- Added dependency [`fast-json-stable-stringify@^2.1.0` ↗︎](https://www.npmjs.com/package/fast-json-stable-stringify/v/2.1.0) (to `dependencies`)
+- Added dependency [`js-sha1@^0.6.0` ↗︎](https://www.npmjs.com/package/js-sha1/v/0.6.0) (to `dependencies`)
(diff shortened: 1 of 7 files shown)


.github/workflows/pr.yml
```diff
@@ -5,14 +5,14 @@ on:
...
 jobs:
-  dependencies:
-    uses: the-guild-org/shared-config/.github/workflows/changesets-dependencies.yaml@main
-    with:
-      installDependencies: true
-      packageManager: yarn
-      nodeVersion: 18
-    secrets:
-      githubToken: ${{ secrets.GUILD_BOT_TOKEN }}
+#   dependencies:
```
ardatan: :)))

n1ru4l: temporary until @dotansimha resolves it 😃

PR #1606, opened 2023-01-04
fix(newrelic): handle invalid newrelic env vars correctly and add tests

ardatan:
I'll add tests for resolver tracking later

### DIFF
--- a/packages/plugins/newrelic/src/index.ts
+++ b/packages/plugins/newrelic/src/index.ts
@@ -1,15 +1,9 @@
-import {
-  ExecutionResult,
-  FieldNode,
-  GraphQLError,
-  Kind,
-  OperationDefinitionNode,
-  print,
-} from 'graphql';
+import { ExecutionResult, FieldNode, getOperationAST, GraphQLError, Kind, print } from 'graphql';
+import newRelic from 'newrelic';
 import { DefaultContext, getDocumentString, isAsyncIterable, Path, Plugin } from '@envelop/core';
 import { useOnResolve } from '@envelop/on-resolve';
 
-enum AttributeName {
+export enum AttributeName {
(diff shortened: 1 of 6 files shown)


saihaj (approved):

saihaj:
Is there a way we can add tests for this?

ardatan:
@saihaj I think so;
https://github.com/newrelic/node-test-utilities
Hmm maybe next sprint I can take a look at this.

packages/plugins/newrelic/src/index.ts
```diff
@@ -125,50 +119,75 @@ export const useNewRelic = (rawOptions?: UseNewRelicOptions): Plugin => {
...
+          ) as FieldNode[];
+          rootFields = fieldNodes.map(fieldNode => fieldNode.name.value);
+        }
+
+        const operationType = rootOperation.operation;
+        const operationName =
+          options.extractOperationName?.(args.contextValue) ||
+          args.operationName ||
+          rootOperation.name?.value ||
+          AttributeName.ANONYMOUS_OPERATION;
```
YassinEldeeb: I think this variable has been defined at the top with the same definition and is accessible from line 140
```suggestion
```

YassinEldeeb (changes requested):
LGTM, besides removing the repeated `operationName` variable

.github/workflows/tests.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         run: yarn test
         env:
           CI: true
+      - name: Test NewRelic
```
saihaj: Why add a new command. Our CI run tests recursively for each package?

ardatan: No, `yarn test` runs the `test` script defined in the root `package.json`. And that script runs `jest` only.

PR #1629, opened 2023-02-06
Fix: graphql-modules did not cleanup on subscription end

### LINKED ISSUE #1628: onExecute hooks not called for subscriptions?
### Issue workflow progress

_Progress of the issue based on the [Contributor Workflow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow)_

- [ ] 1. The issue provides a [minimal reproduction](https://en.wikipedia.org/wiki/Minimal_reproducible_example) available on [Stackblitz](https://stackblitz.com/fork/node).
- [ ] 2. A failing test has been provided
- [ ] 3. A local solution has been provided
- [ ] 4. A pull request is pending review

---

**Describe the bug**

Our application is leaking memory, and profiling reveals that `executionContextDependencyStore` in GraphQL Modules is growing for every new subscription. In our case, there are many subscriptions where a connection is established with subscriptions, and the subscription (separately or along with connection) is closed without ever executing anything.

When checking the implementation of [graphql modules plugin](https://github.com/n1ru4l/envelop/blob/4eaaf76d418cf56476d57c176ca876fff89a86b2/packages/plugins/graphql-modules/src/index.ts),  it appears that related context cleanup is done in onExecute and onExecuteDone hooks.

In our described case, it appears onExecute is never called. The actual setup is very similar to [graphql-ws example](https://github.com/n1ru4l/envelop/blob/dc1e24b5340ed7eba300a702b17f9be5cff65a8f/examples/graphql-ws/index.ts).

Before providing more exact steps and a canned test case, may I ask is this expected behavior, or have I simply broked our envelop so that execute() is never called? In case of subscriptions, how might I cleanup the graphql-module global map, e.g. as part of onComplete() callback from web socket server?

**To Reproduce**

**Expected behavior**

**Environment:**

- OS: Alpine Linux / MacOS
- NodeJS: 18.x
- `@envelop/*` versions: 2.4.x
  - `@envelop/core`: 2.4.2
  - `@envelop/graphql-modules`: 3.4.2

**Additional context**

laurisvan:
## Description

This change replicates a similar cleanup procedure to onSubscribe() that onExecute has in graphql-modules plugin. This was done once we discovered a memory leak in our products related to module injectors not being cleaned up when subscriptions/websocket connections are cleaned up.

The change is almost a copy-paste from the similar onExecute function.

Fixes #1628 

## Type of change

## How Has This Been Tested?

We are running the same change patched with patch-package and verified the memory leak of extraneous injectors has been resolved. A minimal test was added that the code does execute, but I do admit it does not verify the module private variable has been cleaned up. Frankly speaking, I do not know how to best verify that without extensive mocking.

- [x] Created a minimal test to verify subscriptions will not crash despite the cleanup procedure added

**Test Environment**:

- OS: OS/X
- `@envelop/.graphql-modules`:
- NodeJS: 18.0.0

## Checklist:

### DIFF
--- a/.changeset/olive-crabs-accept.md
+++ b/.changeset/olive-crabs-accept.md
@@ -1,0 +1,5 @@
+---
+'@envelop/graphql-modules': patch
+---
+
+Cleanup reference to context value on subscription completion
--- a/packages/plugins/graphql-modules/src/index.ts
+++ b/packages/plugins/graphql-modules/src/index.ts
@@ -29,5 +29,15 @@ export const useGraphQLModules = (app: Application): Plugin => {
         },
       };
     },
+    onSubscribe({ args }) {
+      return {
+        onSubscribeResult: () => {
+          if (args.contextValue && args.contextValue[graphqlModulesControllerSymbol]) {
+            args.contextValue[graphqlModulesControllerSymbol].destroy();
+            args.contextValue[graphqlModulesControllerSymbol] = null;
+          }
+        },
+      };
+    },
   };
 };
--- a/packages/plugins/graphql-modules/test/use-graphql-modules.spec.ts
+++ b/packages/plugins/graphql-modules/test/use-graphql-modules.spec.ts
@@ -1,7 +1,12 @@
 import 'reflect-metadata';
 import { parse } from 'graphql';
 import { Application, createApplication, createModule, Injectable, Scope } from 'graphql-modules';
-import { assertSingleExecutionValue, createTestkit } from '@envelop/testing';
+import {
+  assertSingleExecutionValue,
+  assertStreamExecutionValue,
+  collectAsyncIteratorValues,
+  createTestkit,
+} from '@envelop/testing';
 import { useGraphQLModules } from '../src/index.js';
 
 describe('useGraphQLModules', () => {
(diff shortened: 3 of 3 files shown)


ardatan:
Could you create changesets with `yarn changeset`? Thanks!

laurisvan:
@ardatan should be there now - thanks for the guidance!

n1ru4l:
You can implement a `onDestory` function on the provider that should be invoked. I quickly tested it here: [n1ru4l/envelop@`41d6964` (#1630)](https://github.com/n1ru4l/envelop/pull/1630/commits/41d6964003725ceb314fdf26ff74c945e5d39973)

Seems like there is an issue within graphql-modules. I pinged @kamilkisiela and he will have a look at it today.

laurisvan:
Thanks for spotting onSubscribeError - BTW don't we need onExecuteError as well? Actually this onSubscribeError puzzles me a bit as the docs on the web site don't mention such a hook.

Regarding onDestroy - I didn't quite get it - i read onDestroy should be called only for operation scoped providers. Would that be within the responsibility of this plugin (I don't know how graphql-modules does it without envelop, so I would assume it is already there?)

n1ru4l:
@laurisvan `onExecuteDone` is invoked after `onExecuteError`. I agree that the situation is a bit confusing, we have yet not found the best way of visualizing all the hook phases, sorry for the confusion! If you have any ideas we would appreciate any community contribution.

The reason why we diverged from using the modules provided `execute` and `subscribe` function, that have the cleanup built-in is the following:

> we don’t need to use the custom execute/subscribe of graphql modules. if we’ll use it, it means you can’t combine it with other plugins that overrides that (and have a good reason to)

Let me know if you have any further questions!

PR #1632, opened 2023-02-08
feat(sentry): use WeakMap cache for printed documents

jeengbe:
🚨 **IMPORTANT: Please do not create a Pull Request without creating an issue first.**   welll  ¯\_(ツ)_/¯ why would you discourage MRs

_Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of the pull request._

## Description

In scenarios where the used document object is cached, and therefore the same by reference, an option is provided to opt in to an LRU cache so that the document doesn't need to be reparsed over and over.

Fixes # (issue)

## Type of change

n/a

## How Has This Been Tested?

Unit tests should pass

**Test Environment**:

n/a

## Checklist:

## Further comments

n/a

### DIFF
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -98,7 +98,7 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
   const skipOperation = pick('skip', () => false);
   const skipError = pick('skipError', defaultSkipError);
 
-  const printedDocumentsCache = new LRU<string, string>({ max: DEFAULT_MAX, maxAge: DEFAULT_TTL });
+  const printedDocumentsCache = new LRU<any, string>({ max: DEFAULT_MAX, maxAge: DEFAULT_TTL });
 
   const eventIdKey = options.eventIdKey === null ? null : 'sentryEventId';
 
--- a/packages/plugins/sentry/src/index.ts
+++ b/packages/plugins/sentry/src/index.ts
@@ -9,7 +9,6 @@ import type { Span, TraceparentData } from '@sentry/types';
 import { ExecutionArgs, GraphQLError, Kind, OperationDefinitionNode, print } from 'graphql';
 import LRU from 'lru-cache';
 
-
 export type SentryPluginOptions = {
   /**
    * Starts a new transaction for every GraphQL Operation.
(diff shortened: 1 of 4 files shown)


packages/plugins/sentry/src/index.ts
```diff
@@ -92,6 +104,12 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
   const renameTransaction = pick('renameTransaction', false);
   const skipOperation = pick('skip', () => false);
   const skipError = pick('skipError', defaultSkipError);
+  const cachePrintedDocuments = pick('cachePrintedDocuments', false);
```
ardatan: I think caching the document from `onParse` might be better.

jeengbe: What do you mean? This MR is about caching the "expensive" `print()`, not any parsing.

ardatan:
```ts
const documentSDLMap = new WeakMap<DocumentNode, string>();
{
  onParse({ params }) {
      return ({ result }) => {
         documentSDLMap.set(result, params.source.toString());
      }
  },
  onExecute({ args }) {
    // Just in case `parse` is not called by the server, fallback to `print`
    const sdl = documentSDLMap.get(args.document) || print(sdl);
  }
}
```

jeengbe: The idea is good, but sadly this is not quite the same, as the result from `print()` is formatted whereas this could be any raw minified input sdl.

jeengbe: Also, not exactly what I was aiming for in this MR, but thinking about it, a WeakMap might be the right choice after all. I initially decided against it due to its uselessness with freshly parsed documents. 

My use case for this MR is that with a persisted operations plugin, the documents object is always the same by reference for most queries, which is a golden opportunity for caching.

ardatan: Persisted operations plugin also set the `DocumentNode` so the same object is used then it works exactly like you said if I don't miss anything.

jeengbe: Yeah, but my point is that you can't use `params.source.toString()` since that's most likely a non-indented blob input, so there needs to be at least one call to `print()`, which might as well be this one.

ardatan (approved):

packages/plugins/sentry/src/index.ts
```diff
@@ -157,7 +165,7 @@ export const useSentry = (options: SentryPluginOptions = {}): Plugin => {
           // eslint-disable-next-line no-console
           console.warn(
             [
-              `Flag "startTransaction" is enabled but Sentry failed to find a transaction.`,
+              `Flag "startTransaction" is disabled but Sentry failed to find a transaction.`,
```
ardatan, line 168: I think this is not related :)

ardatan (changes requested):

packages/plugins/sentry/src/index.ts
jeengbe: It's not, I just noticed it and thought I'd fix it while I'm at it :]

n1ru4l (approved):

PR #1723, opened 2023-02-14
revert guild/components which causes site to crash

n1ru4l:


dotansimha:
@n1ru4l I think the upgrade came with some API changes @B2o5T did? that's why build is failing now?

PR #1731, opened 2023-02-21
Update components + remove v2 folder, fetch v2 docs from git

dimaMachina:


dotansimha (approved):

n1ru4l:
@B2o5T can we use https://github.com/n1ru4l/envelop/tree/v2 instead of the exact commit and also add an edit this page on github button? Could we also automatically inject an "maybe you should use latest version" info box into each of these v2 pages?

n1ru4l (approved):

PR #1758, opened 2023-04-24
use pnpm instead of ~~npm~~ yarn

n1ru4l:
if you read this you are fired


kamilkisiela:
> **Conflicting files**
> yarn.lock

I love it

.github/workflows/benchmark.yml
```diff
@@ -31,10 +31,10 @@ jobs:
             ${{ runner.os }}-17-16-yarn-
 
       - name: Install Dependencies using Yarn
```
dotansimha: @n1ru4l you can replace this whole section (node, cache, pnpm, pnpm i) with the setup action from shared-config.

dotansimha: also in the other pipelines

package.json
dotansimha, line 1: 1. Remove `workspaces` from this file
2. Replace `resolutions` with the `pnpm`-specific thing?

pnpm-lock.yaml
dotansimha, line 1: if this was created from scratch, it might have "newer" dependencies somewhere. if you wish to avoid that and have the same as `yarn.lock`, you can use https://pnpm.io/cli/import

PR #1776, opened 2023-05-22
Drop Node 14 and require Node 16 or higher

ardatan:


.github/workflows/pr.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     with:
       installDependencies: true
       packageManager: pnpm
-      nodeVersion: 18
+      nodeVersion: 20
```
n1ru4l: lets use the LTS version here and in all other sections

ardatan: 👍

n1ru4l:
tsconfig target and lib can be bumped to es2021, see https://github.com/kamilkisiela/graphql-config/blob/master/tsconfig.json#L4-L5

ardatan:
Done @n1ru4l 👍

dotansimha (approved):

PR #1896, opened 2023-06-25
feat(response-cache): support defer and stream

### LINKED ISSUE #1122: response cache with defer/stream
It should be possible to cache results that use defer and stream

EmrysMyrddin:
## Description

The reponse-cache plugin should handle queries using `@defer` or `@stream` directives.

Fixes #1122 

## Type of change

## Checklist:

- [x] I have followed the

## Further comments

For now, I've made the choice to merge all incremental patches and serve the result on cache hit. An other approach would be to save the patches and replay them on cache hit.
I've this choice because anyway, with the current plugin implementation, I need the merged result to calculate if the result should be cached and the appropriate TTL.

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -13,7 +13,6 @@ import {
 import {
   ExecutionResult,
   getDocumentString,
-  IncrementalExecutionResult,
   isAsyncIterable,
   Maybe,
   ObjMap,
@@ -516,8 +515,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           let result: ExecutionResult = {};
           return {
             onNext(payload) {
-              const { data, errors, extensions, incremental, hasNext } =
-                payload.result as IncrementalExecutionResult;
+              const { data, errors, extensions } = payload.result;
 
               if (data) {
                 // This is the first result with the initial data payload sent to the client. We use it as the base result
@@ -532,15 +530,19 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
                 }
               }
 
-              if (incremental) {
-                for (const patch of incremental) {
-                  mergeIncrementalResult(result, patch);
-                }
-              }
+              if ('hasNext' in payload.result) {
+                const { incremental, hasNext } = payload.result;
 
-              if (!hasNext) {
-                // The query is complete, we can process the final result
-                maybeCacheResult(result, payload.setResult);
+                if (incremental) {
+                  for (const patch of incremental) {
+                    mergeIncrementalResult(result, patch);
+                  }
+                }
+
+                if (!hasNext) {
+                  // The query is complete, we can process the final result
+                  maybeCacheResult(result, payload.setResult);
+                }
               }
             },
           };
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -13,6 +13,7 @@ import {
 import {
   ExecutionResult,
   getDocumentString,
+  IncrementalExecutionResult,
   isAsyncIterable,
   Maybe,
   ObjMap,
@@ -508,7 +509,10 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           // The merged result is then used to know if we should cache it and to calculate the ttl.
           let result: ExecutionResult = {};
           return {
-            onNext({ result: { data, incremental, hasNext, errors, extensions }, setResult }) {
+            onNext(payload) {
+              const { data, errors, extensions, incremental, hasNext } =
+                payload.result as IncrementalExecutionResult;
+
               if (data) {
                 // This is the first result with the initial data payload sent to the client. We use it as the base result
                 if (data) {
@@ -530,7 +534,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
 
               if (!hasNext) {
                 // The query is complete, we can process the final result
-                maybeCacheResult(result, setResult);
+                maybeCacheResult(result, payload.setResult);
               }
             },
           };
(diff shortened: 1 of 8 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -441,42 +442,76 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
-          cache.set(cacheKey, processedResult, identifier.values(), finalTtl);
-          if (includeExtensionMetadata) {
-            setResult(
-              resultWithMetadata(processedResult, { hit: false, didCache: true, ttl: finalTtl }),
-            );
-          }
+          // When the result is an AsyncIterable, it means the query is using @defer or @stream.
+          // This means we have to build the final result by merging the incremental results.
+          // The merged result is then used to know if we should cache it and to calculate the ttl.
+          let result: ExecutionResult = {};
```
EmrysMyrddin, line 487: I'm not happy with this, if someone knows how to make typescript happy by discriminating the union type... it seems `if(result.incremental)` is not working, while I thought it was the way to do it.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -441,42 +443,79 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
+          return {
+            onNext(payload) {
+              const { data, errors, extensions, incremental, hasNext } =
```
EmrysMyrddin: I'm not happy with this, if someone knows how to make typescript happy by discriminating the union type... it seems `if(payload.result.incremental)` is not working, while I thought it was the way to do it.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -509,3 +548,51 @@ function calculateTtl(typeTtl: number, currentTtl: number | undefined): number {
...
+  // We need to follow the path to the point where the patch should be applied
+  let target: any = result;
+  const path = ['data', ...(patch.path ?? [])];
+  for (let i = 0; i < path.length - 1; i++) {
+    target = target[path[i]];
+  }
+  const prop = path[path.length - 1];
+
+  const items = (patch as IncrementalStreamResult).items;
+  const data = (patch as IncrementalDeferResult).data;
```
EmrysMyrddin: And I have the same Typescript problem here, where I can't discriminate the union type.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -13,11 +13,13 @@ import {
 import {
   ExecutionResult,
   getDocumentString,
+  IncrementalExecutionResult,
   isAsyncIterable,
   Maybe,
   ObjMap,
   Plugin,
 } from '@envelop/core';
+import { IncrementalDeferResult, IncrementalStreamResult } from '@graphql-tools/executor';
```
ardatan: Hmm if we really need these types, we can move them to `utils` package to avoid adding `executor` as a dependency.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -509,3 +548,51 @@ function calculateTtl(typeTtl: number, currentTtl: number | undefined): number {
 export const cacheControlDirective = /* GraphQL */ `
   directive @cacheControl(maxAge: Int) on FIELD_DEFINITION | OBJECT
 `;
+
+function mergePatch(
```
ardatan: Do you think we can use this function instead?
https://github.com/ardatan/graphql-tools/blob/master/packages/executors/http/src/handleMultipartMixedResponse.ts#L52

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: Oh yeah I didn't realized they were part of `executor`, I've added it for the tests, didn't meant to add it as a real dependency ><
So you think it is ok to just copy/paste it ?

ardatan: Do we really need that strict types for these two? Aren't they already inside `ExecutionResult` of `utils` package?

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: Yes it seems to do pretty much the same thing, except it doesn't merge the extensions.
Perhaps I can make a PR to add extension merging to the `@graphql-tools/utils` function, since I think it should do it.

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: Oh yes no I understood, you mean move it from `@graphql-tools/executor` to `@graphql-tools/utils`

packages/plugins/response-cache/src/plugin.ts
ardatan: Sure @EmrysMyrddin ! Go for it!

EmrysMyrddin:
I will make a PR on `graphql-tools` to make the required changes.

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: TS complains about `incremental` and `hasNext` not being declared on `ExecutionResult` ><. From the types, it seems `ExecutionResult` doesn't include `IncrementalResult`, which is true, you can't have an `IncrementalResult` as a direct result of execution.

EmrysMyrddin: I just reviewed my code, and there is something I don't understand. In fact, we don't use `@graphql-tools/utils` types, but our own redefined types inside envelop. So I updated (but it seems I've failed to use it ^^') the envelop's types https://github.com/n1ru4l/envelop/pull/1896/files#diff-03f876d59dce271fae6bf2c434d7a111a2ebd06c787c73209dca96008bee8e23

EmrysMyrddin: Actually we can't use graphql tools type since envelop core doesn't use it. I have managed to not use a type assertion, I'm not sure it's better ^^

ardatan (approved):

ardatan (approved):

PR #1935, opened 2023-08-31
upgrade newrelic

EmrysMyrddin:
## Description

Update peer range for NewRelic plugin.

## Type of change

Dependency update

### DIFF
--- a/.changeset/@envelop_newrelic-1935-dependencies.md
+++ b/.changeset/@envelop_newrelic-1935-dependencies.md
@@ -1,0 +1,8 @@
+---
+'@envelop/newrelic': patch
+---
+
+dependencies updates:
+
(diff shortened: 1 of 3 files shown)


dotansimha (approved):

.changeset/@envelop_newrelic-1935-dependencies.md
```diff
@@ -0,0 +1,8 @@
+---
+'@envelop/newrelic': patch
```
dotansimha, line 2: @EmrysMyrddin this is really a patch? can we verify it doesn't require major?

ardatan: We don't drop the support for older versions, so I think patch is fine.

EmrysMyrddin: Yea, it's why I updated manually instead of merging renovate PR :-)

PR #1939, opened 2023-09-05
better plugin order when using addPlugin API

EmrysMyrddin:
## Description

In Envelop, plugin hooks execution order is most of the time very important. Some plugins can depend on some data a previous plugin put in the context to work properly for example.

Today, the execution order can be a bit tricky to predict because if a plugin rely under the hood on `addPlugin` API, it can push new hooks to the end of the plugin list. It is particularly visible with the `useEnvelop` which insert its plugins at the end of the plugin list, which is a bit confusing and difficult to predict for most of developers.

Here a basic example:
```ts
import baseEnvelop from './shared/envelop'
import useLogger from './logger-plugin'

  plugins: [
    useEnvelop(baseEnvelop), // <= Can break the logger since it will execute after it while being defined before it
    useLogger(),
  ]
})
```

## Type of change

      expected)

> Since this introduces slight difference in execution order, some things can break in existing code, but it should be rare and visible very quickly.

## How Has This Been Tested?

It passes all the current test suites

## Further comments

This comes in response to multiple issues or support messages we got concerning this exact problem of unexpected execution order.

### DIFF
--- a/.changeset/thick-walls-agree.md
+++ b/.changeset/thick-walls-agree.md
@@ -1,0 +1,12 @@
+---
+'@envelop/core': major
+---
+
+The `addPlugin` function now insert the plugin in place in the plugin list, leading to a more
+predictable execution order.
+
+**Breaking Change:** This change alter the execution order of plugins. This can break some plugins
+that was relying on the fact the `addPlugin` allowed to push a plugin to the end of the plugin list.
+
+If it is the case, the best fix is to reorder the plugin list and ensure the plugin is in the right
+position, after all its dependencies.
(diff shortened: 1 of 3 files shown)


ardatan (approved):

ardatan:
Do you think we can add tests to prevent future regressions?

EmrysMyrddin:
@ardatan Yes sure :-) It was more a POC to discuss about it but if you think it's ok, I can write tests !

I have to add a changeset, do you think we need to release a major for this ?

ardatan:
> I have to add a changeset, do you think we need to release a major for this ?

It is more like a fix. I want to hear @n1ru4l and @dotansimha 's opinions on this.

EmrysMyrddin:
I agree that in the spirit it's a fix, but since it will introduce a breaking change... to really follow semver perhaps we will have to release a major. Perhaps we can wait to have more breaking features before releasing this ?

n1ru4l (approved):
I agree that we should release this as a breaking change.

PR #1948, opened 2023-09-20
ignore subscriptions in response cache plugin

EmrysMyrddin:
## Description

Response cache plugin adds type names to every selection set in the query documents to allow introspection and TTL calculation. This introduce a validation error when a subscription is used since the root subscription selection set should only contain one field. Since response cache has no sense in the context of subscriptions, it is safe to just ignore subscription documents.

Fixes https://github.com/dotansimha/graphql-yoga/issues/3005

## Type of change

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -209,15 +209,15 @@ const originalDocumentMap = new WeakMap<DocumentNode, DocumentNode>();
 const addTypeNameToDocument = memoize1(function addTypeNameToDocument(
   document: DocumentNode,
 ): DocumentNode {
-  if (
-    document.definitions.some(
-      def => def.kind === Kind.OPERATION_DEFINITION && def.operation === 'subscription',
-    )
-  ) {
-    return document;
-  }
   let documentChanged = false;
   const newDocument = visit(document, {
+    OperationDefinition(node) {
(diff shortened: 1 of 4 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -209,6 +209,13 @@ const originalDocumentMap = new WeakMap<DocumentNode, DocumentNode>();
 const addTypeNameToDocument = memoize1(function addTypeNameToDocument(
   document: DocumentNode,
 ): DocumentNode {
+  if (
+    document.definitions.some(
+      def => def.kind === Kind.OPERATION_DEFINITION && def.operation === 'subscription',
```
ardatan: What if a document contains multiple subscriptions but a query is selected?

EmrysMyrddin: I didn't though this was even a thing ><

How does it works ? It will both run the execution and subscription in parallel ? 

I should then probably just filter definitions and modify only queries and mutations.

EmrysMyrddin: Ok, I think i've understood how it's supposed to work. I had to modify the test kit to be able to give it the operation name.

ardatan:
Fixed by https://github.com/n1ru4l/envelop/pull/1956

Issue #1953, opened 2023-09-25
@envelop/response-cache clarify handling of null or empty values


marhaupe:
Recently we've seen some issues with stale data coming from our GraphQL server. It's pretty green field so the general entities involved could be anything, so let's just say we have `Product`, `Query.product` and `Mutation.createProduct`:

```graphql
type Query {
  products: [Product!]!
}

type Mutation {
  addProduct(input: ProductInput!): Product!
}

input ProductInput {
  name: String!
}

type Product {
  id: ID!
  name: String!
}
```

Whenever we handle `createProduct`, we invoke `cache.invalidate([typename: "Product"])`, so every operation related to `Product` would get invalidated - or so we thought. Turns out this only invalidates operations where a `Product` entity has been included in the response. If `Query.products` at some point returned an empty list, the response would not get invalidated on a subsequent `Mutation.createProduct` invocation. An exemplary test in `response-cache.spec.ts` should clarify:

```js
  it('purges cached empty query operation execution result via imperative cache.invalidate api using typename', async () => {
    const spy = jest.fn(() => []);
    let id = 1;
    const schema = makeExecutableSchema({
      typeDefs: /* GraphQL */ `
        type Query {
          products: [Product!]!
        }

        type Mutation {
          addProduct(input: ProductInput!): Product!
        }

        input ProductInput {
          name: String!
        }

        type Product {
          id: ID!
          name: String!
        }
      `,
      resolvers: {
        Query: {
          products: spy,
        },
        Mutation: {
          addProduct(_, { input }) {
            const mockProduct = {
              id: id++,
              name: input.name,
            };
            return mockProduct;
          },
        },
      },
    });

    const cache = createInMemoryCache();
    const testInstance = createTestkit([useResponseCache({ session: () => null, cache })], schema);

    const query = /* GraphQL */ `
      query test {
        products {
          id
          name
        }
      }
    `;

    await testInstance.execute(query);
    await testInstance.execute(query);
    expect(spy).toHaveBeenCalledTimes(1);

    cache.invalidate([{ typename: 'Product' }]);

    await testInstance.execute(query);
    expect(spy).toHaveBeenCalledTimes(2);
  });
```
```
  ● useResponseCache › purges cached empty query operation execution result via imperative cache.invalidate api using typename

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 2
    Received number of calls: 1

      532 |
      533 |     await testInstance.execute(query);
[...]

PR #1956, opened 2023-09-25
enhance(response-cache): do not add __typename to the root of subscription operations

ardatan:


n1ru4l:
@ardatan I think we can skip running the logic for processing the document for GraphQL subscriptions? Only query operations need `__typename`. Mutation documents only need  `__typename` if automatic invalidation is enabled 🤔

ardatan:
@n1ru4l Good point! Adding it!

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -206,38 +209,40 @@ export type ResponseCacheExecutionResult = ExecutionResult<
...
-                value: '__typename',
-              },
-            },
-            ...node.selections,
-          ],
-        };
+    SelectionSet(node, _key, parent) {
+      if (parent && 'kind' in parent && parent.kind === Kind.OPERATION_DEFINITION) {
+        if (parent.operation === 'mutation' && !addTypeNameToDocumentOpts.invalidateViaMutation) {
+          return BREAK;
```
EmrysMyrddin: You probably want to return `false` here ? If the mutation invalidation is disabled, we can skip the entire subtree of the AST

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -320,90 +327,45 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
+          }
+          for (const fieldName in data) {
+            if (scopePerSchemaCoordinate[`${typename}.${fieldName}`] === 'PRIVATE' && !sessionId) {
+              skip = true;
+            }
+            if (!skip) {
+              const id = data[fieldName];
+              if (idFields.includes(fieldName)) {
+                identifier.set(`${typename}:${id}`, { typename, id });
+              }
```
EmrysMyrddin: We probably can move the id lookup in the condition ?
```suggestion
              if (idFields.includes(fieldName)) {
                const id = data[fieldName];
                identifier.set(`${typename}:${id}`, { typename, id });
              }
```

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -320,90 +327,45 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
+      function processResult(data: any) {
+        if (data == null || typeof data !== 'object') {
+          return;
+        }
+        if (Array.isArray(data)) {
+          for (const item of data) {
+            processResult(item);
+          }
+          return;
+        }
```
EmrysMyrddin, line 349: It would be great to find a way to handle this case by adding the typename to the collected entities. This would allow to keep track of cached responses using some types but with empty list or null result.

Today, those response doesn't get invalidated when using `cache.invalidate('Type')`, leading to stale cache hits.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -206,38 +209,40 @@ export type ResponseCacheExecutionResult = ExecutionResult<
...
-            ...node.selections,
-          ],
-        };
+    SelectionSet(node, _key, parent) {
+      if (parent && 'kind' in parent && parent.kind === Kind.OPERATION_DEFINITION) {
+        if (parent.operation === 'mutation' && !addTypeNameToDocumentOpts.invalidateViaMutation) {
+          return BREAK;
+        }
+        if (parent.operation === 'subscription') {
+          return addTypeNameToDocumentOpts.invalidateViaSubscription ? node : BREAK;
```
EmrysMyrddin: Same here, I'm not sure `BREAK` is the right thing ? Doesn't it just stop the visitor there ? I though skipping the node subtree was done by returning `false`

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: But perhaps this can be part of a subsequent PR if you think it's out of scope

EmrysMyrddin: Just for reference, here is an issue about it (with a test case!) https://github.com/n1ru4l/envelop/issues/1953

packages/plugins/response-cache/src/plugin.ts
ardatan: Good catch! Thanks!

packages/plugins/response-cache/src/plugin.ts
ardatan: Good catch! Thanks!

packages/plugins/response-cache/src/plugin.ts
ardatan: Let's track that in another PR/issue.

PR #1961, opened 2023-09-27
fix cache invalidations when result contains empty values

### LINKED ISSUE #1953: @envelop/response-cache clarify handling of null or empty values

marhaupe:
🚨 **IMPORTANT: Please do not create a Pull Request without creating an issue first.**

the pull request._

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant
motivation and context. List any dependencies that are required for this change.

Fixes #1953 (issue)

## Type of change

## How Has This Been Tested?

Issue #1953 contains most relevant information, also I added a new test `purges cached empty query operation execution result via imperative cache.invalidate api using typename`.

**Test Environment**:

- OS: MacOS
- `@envelop/...`:
- NodeJS: 18.16.0

## Checklist:

- [x] I have followed the
- [ ] I have commented on my code, particularly in hard-to-understand areas

## Further comments

This might as well just be a draft. I'm not sure if I need to follow the `CONTRIBUTING.md` for drafts. Let me know if this is useful or not.

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -278,6 +278,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
 }: UseResponseCacheParameter<PluginContext>): Plugin<PluginContext> {
   const ignoredTypesMap = new Set<string>(ignoredTypes);
   const processedSchemas = new WeakSet();
+  const typePerSchemaCoordinateMap = new Map<string, string[]>();
 
   // never cache Introspections
   ttlPerSchemaCoordinate = { 'Query.__schema': 0, ...ttlPerSchemaCoordinate };
@@ -295,10 +296,9 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
       if (processedSchemas.has(schema)) {
         return;
       }
-      // Check if the schema has @cacheControl directive
       const cacheControlDirective = schema.getDirective('cacheControl');
-      if (cacheControlDirective) {
-        mapSchema(schema, {
+      mapSchema(schema, {
+        ...(cacheControlDirective && {
           [MapperKind.COMPOSITE_TYPE]: type => {
             const cacheControlAnnotations = getDirective(schema, type, 'cacheControl');
             cacheControlAnnotations?.forEach(cacheControl => {
@@ -312,7 +312,12 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
             });
             return type;
           },
-          [MapperKind.FIELD]: (fieldConfig, fieldName, typeName) => {
+        }),
+        [MapperKind.FIELD]: (fieldConfig, fieldName, typeName) => {
+          const schemaCoordinates = `${typeName}.${fieldName}`;
+          const resultTypeNames = unwrapTypenames(fieldConfig.type);
+          typePerSchemaCoordinateMap.set(schemaCoordinates, resultTypeNames);
+          if (cacheControlDirective) {
             const cacheControlAnnotations = getDirective(schema, fieldConfig, 'cacheControl');
             cacheControlAnnotations?.forEach(cacheControl => {
               const ttl = cacheControl.maxAge * 1000;
@@ -323,10 +328,10 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
                 scopePerSchemaCoordinate[`${typeName}.${fieldName}`] = cacheControl.scope;
               }
             });
-            return fieldConfig;
-          },
-        });
-      }
+          }
+          return fieldConfig;
+        },
+      });
       processedSchemas.add(schema);
     },
     async onExecute(onExecuteParams) {
@@ -363,6 +368,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
             currentTtl = calculateTtl(ttlPerType[typename], currentTtl);
           }
           for (const fieldName in data) {
+            const fieldData = data[fieldName];
             if (!skip) {
               if (
                 scopePerSchemaCoordinate[`${typename}.${fieldName}`] === 'PRIVATE' &&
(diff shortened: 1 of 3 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -295,11 +296,10 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
       if (processedSchemas.has(schema)) {
         return;
       }
-      // Check if the schema has @cacheControl directive
       const cacheControlDirective = schema.getDirective('cacheControl');
-      if (cacheControlDirective) {
```
ardatan, line 300: Is this really needed because this visits over the schema even if `cacheControl` is not present.

marhaupe: Yeah this is by no means ready to merge, I honestly did not want to risk optimizing a solution that I'm not sure is actually wanted. If we're already talking about it though, what would you suggest? I would probably try to iterate over the schema twice to keep things clean. Keep the handling for `@cacheControl` as it is, and generate the `Map` in a separate iteration separately, depending on `invalidateViaMutations`

EmrysMyrddin:
Thank you for sharing this with us !

Honestly, I'm pretty happy with your proposal. I think that it is a good improvement, the current behaviour of the cache invalidation for empty values is actually a bug we need to fix I think.

Your approach seems totally fine for me !

I will add some observations but I think this is totally something we want to merge :-)

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -310,9 +310,14 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
               }
             });
-            return type;
-          },
-          [MapperKind.FIELD]: (fieldConfig, fieldName, typeName) => {
+          }
+          return type;
+        },
+        [MapperKind.FIELD]: (fieldConfig, fieldName, typeName) => {
+          const id = `${typeName}.${fieldName}`;
```
EmrysMyrddin: Perhaps call this variable `schemaCoordinates` ?
```suggestion
          const schemaCoordinates = `${typeName}.${fieldName}`;
```

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -480,6 +480,66 @@ describe('useResponseCache', () => {
     expect(spy).toHaveBeenCalledTimes(2);
   });
 
+  it('purges cached empty query operation execution result via imperative cache.invalidate api using typename', async () => {
```
EmrysMyrddin: Can you add a test for the null case too ?

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: I think that it's ok like this. If you want to make things more clear, you can eventually create functions with meaningful names to make it clear what is done in each visitor hook. But visiting the entire schema twice can hurt initialisation performance and hurt the startup time :-)

EmrysMyrddin: And I think we should always build the Map, even when automatic cache invalidation is turned off, since we still want to be able to manually invalidate the cache.

EmrysMyrddin (changes requested):

marhaupe:
Thanks for your quick feedback! I've just pushed some changes, pretty happy with how it's turned out.

packages/plugins/response-cache/src/plugin.ts
marhaupe: Right now I'm not visiting `MapperKind.COMPOSITE_TYPE` if `cacheControlDirective` does not exist. I think this is as good as it gets for now.

EmrysMyrddin:
Thank you for taking the time for this 😀 

After some thoughts, I see a drawback with our approach here.

The idea here is to be able to invalidate by type when we add an entity to avoid having null or lists stale. But this comes at cost of loosing the cache for any response which contains at least one entity of this type, while it is probably not stale just because we added an item.

It is still fine to merge this, I should probably open a new issue to discuss about this problem, it's probably more a lack of api for the invalidation part of the cache.

EmrysMyrddin (approved):

EmrysMyrddin:
I've created an issue to discuss about my concern here : #1979 

Let's merge this since it's not really related.

@marhaupe Can you add a changeset ? Just run `pnpm changeset` and follow instructions :-)

marhaupe:
Done 👍 I also read through your proposal and I like it, I just left a comment with one thought I had though.

EmrysMyrddin:
Thank you again for the time spent on this :-) very appreciated !

PR #1970, opened 2023-09-29
[response-cache] automatically add entity id to document

### LINKED ISSUE #1969: [response-cache] automatically add id field to the query document
**Describe the bug**

Since the cache rely on entity id to link entity to cached response, the id needs to be part of the selection of each composite type.

We have the same requirement for the `__typeName`, which is automatically added to the document to avoid missing this field.

We should probably do the same thing for the id field.

EmrysMyrddin:
## Description

Since the cache rely on entity id to link entity to cached response, the id needs to be part of the selection of each composite type.

We have the same requirement for the `__typeName`, which is automatically added to the document to avoid missing this field.

This PR aims to do the same thing with the id field.

Fixes #1969

## Type of change

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -1,6 +1,5 @@
 import jsonStableStringify from 'fast-json-stable-stringify';
 import {
-  ASTVisitor,
   DocumentNode,
   ExecutionArgs,
   getOperationAST,
@@ -23,7 +22,7 @@ import {
   getDirective,
   MapperKind,
   mapSchema,
-  memoize2,
+  memoize4,
   mergeIncrementalResult,
 } from '@graphql-tools/utils';
 import type { Cache, CacheEntityRecord } from './cache.js';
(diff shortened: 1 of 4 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -207,39 +208,9 @@ export type ResponseCacheExecutionResult = ExecutionResult<
 const originalDocumentMap = new WeakMap<DocumentNode, DocumentNode>();
 const addTypeNameToDocument = memoize2(function addTypeNameToDocument(
```
EmrysMyrddin: I'm not sure why do we memoize this since we also check if the document has been parsed before calling it.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -304,22 +288,73 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
+                scopePerSchemaCoordinate[schemaCoordinates] = cacheControl.scope;
               }
             });
           }
           return fieldConfig;
         },
       });
-      processedSchemas.add(schema);
+
+      const typeInfo = new TypeInfo(schema);
```
ardatan: I'm not sure if TypeInfo is reusable.

packages/plugins/response-cache/src/plugin.ts
ardatan: What do you mean by checking if it has been parsed?

EmrysMyrddin: I mean that we have this guard at the start of the `onSchemaChange` which prevent the execution of the function :
```ts
      if (processedSchemas.has(schema)) {
        visitor = processedSchemas.get(schema)!;
        return;
      }
```

And the only way this is true is if the function has already been called with this document, with this line at the end of `addTypeNameToDocument:

```ts
	originalDocumentMap.set(newDocument, document);
```

So I'm not sure we ever make a memoized call.

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: In theorie, it is reusable because once you visited all the tree, we come back to the initial empty state. But It's probably not a good idea to rely on the fact the visitor actually traversed the whole tree :-(

This means we have to instantiate a new visitor for each execution :-/

packages/plugins/response-cache/src/plugin.ts
ardatan: If the document is cached by parser cache or sth else, this memoization allows us to avoid multiple visits over the document.
`originalDocumentMap` is a different map that keeps documents in other way around.

packages/plugins/response-cache/src/plugin.ts
ardatan: Reverting this logic;
https://github.com/n1ru4l/envelop/pull/1970/files#diff-cf77bd9608a2b0bf192f49649a89018e1c8d8d171b0ee9f7d3f5eca4e230d656R211
and moving TypeInfo instantiation to the function body should be ok. But in this case, memoization should contain schema as well.

packages/plugins/response-cache/src/plugin.ts
EmrysMyrddin: Hummm I think I understand the memoization part, but not why we check for the presence in the map then. In which situation we could have the document being present as a key in the `originalDocumentMap` if the keys are the modified documents ?

PR #1977, opened 2023-10-02
always eagerly check if cache is enable for the current request

### LINKED ISSUE #1963: `@envelop/response-cache` not respecting `enabled` being `false`
### Issue workflow progress

- [x] 1. The issue provides a minimal reproduction available on [Stackblitz](https://stackblitz.com/edit/node-bat5aq?file=index.js).

---

**Describe the bug**

I am getting cache writes even though enabled is returning false, and debugging things I can see in the source code that `@envelop/response-cache` doesn't consider `enabled()` fn when writing to cache:

```ts
            // THIS IS `false`

            if ((enabled?.(onExecuteParams.args.contextValue) ?? true) === true) {
                const cachedResponse = (await cache.get(cacheKey));
                if (cachedResponse != null) {
                    if (includeExtensionMetadata) {
                        onExecuteParams.setResultAndStopExecution(resultWithMetadata(cachedResponse, { hit: true }));
                    }
                    else {
                        onExecuteParams.setResultAndStopExecution(cachedResponse);
                    }
                    return;
                }
            }
            function maybeCacheResult(result, setResult) {
                processResult(result.data);
                // we only use the global ttl if no currentTtl has been determined.
                const finalTtl = currentTtl ?? globalTtl;
                if (skip || !shouldCacheResult({ cacheKey, result }) || finalTtl === 0) {
                    if (includeExtensionMetadata) {
                        setResult(resultWithMetadata(result, { hit: false, didCache: false }));
                    }
                    return;
                }

                // THIS STILL GETS CALLED

                console.log('maybeCacheResult > cache.set');
                cache.set(cacheKey, result, identifier.values(), finalTtl);
                if (includeExtensionMetadata) {
                    setResult(resultWithMetadata(result, { hit: false, didCache: true, ttl: finalTtl }));
                }
            }
            return {
                onExecuteDone({ result, setResult }) {
                    if (isAsyncIterable(result)) {
                        return handleAsyncIterableResult(maybeCacheResult);
                    }
                    return maybeCacheResult(result, setResult);
                },
            };
```

1. Visit https://stackblitz.com/edit/node-bat5aq?file=index.js
2. Open /graphql
3. Query `query { hello }`
4. BOOM cache.set is called

**Expected behavior**

When enabled is `false` ALL behaviour in this plugin should be skipped. Especially cache.set / cache.get / cache.invalidate!

**Environment:**

See stackblitz

**Additional context**

N/A

EmrysMyrddin:
## Description

The `enabled` parameter allows the user to disabled the cache at a request level based on the context. This give the user fine grain control of the cache behaviour.

But the current implementation only allow to disabled **reading** from the cache. The rest of the pipeline is still executed, the request result will still be written to the cache, and it will still invalidate the cache.

This PR aims to modify this behaviour by using `enabled` to entirely disable the cache. For this, I check as eagerly as possible the `enabled` function to stop the cache related processing as soon as possible.

Fixes #1963 

## Type of change

## More Context

I'm not sure if I should entirely disable the cache, or if we should still always do the automatic invalidation. Or perhaps having 2 options to control this 2 features separately ?

### DIFF
--- a/.changeset/strong-eagles-agree.md
+++ b/.changeset/strong-eagles-agree.md
@@ -1,0 +1,12 @@
+---
+'@envelop/response-cache': major
+---
+
+The `enable` parameter now allows to entirely disable caching. It is checked eagerly and disables
+all cache related processing.
+
+**Breaking Change:**
+
+Previously, `enable` was only controlling cache reading. This means that previously, the automatic
+cache invalidation was still working even with `enable` returning false, which is no longer the
+case. The alternative is to cautiously invalidate data in the related resolvers.
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -22,6 +22,7 @@ import {
   getDirective,
   MapperKind,
   mapSchema,
+  memoize1,
   memoize4,
   mergeIncrementalResult,
 } from '@graphql-tools/utils';
@@ -284,6 +285,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
 }: UseResponseCacheParameter<PluginContext>): Plugin<PluginContext> {
   const ignoredTypesMap = new Set<string>(ignoredTypes);
   const typePerSchemaCoordinateMap = new Map<string, string[]>();
+  enabled = enabled ? memoize1(enabled) : enabled;
 
   // never cache Introspections
   ttlPerSchemaCoordinate = { 'Query.__schema': 0, ...ttlPerSchemaCoordinate };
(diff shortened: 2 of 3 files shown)


EmrysMyrddin:
I'm not sure if this is considered a breaking change and if we should do a major version to release this.

If we think it's breaking a breaking change, we should consider shipping #1939 too.

n1ru4l (changes requested):
Can we also add tests for this, to catch regressions?
Technically, this is also a breaking change, so let's release this as a major version with a changeset.

EmrysMyrddin:
@n1ru4l Do you think the test is sufficient ? I was not sure how to test this, so i just tested that the cache object is not used when `enabled` return false.

ardatan (approved):

Issue #1981, opened 2023-10-04
useOnResolve allow for introspection types


darren-west:
### Problem
I am using the useOnResolve plugin to implement some custom functionality around auth, I wanted to be able to apply some custom logic to introspection but found the plugin excludes this.

### Solution
Provide an option to support introspection types, make it disabled by default so it is backwards compatible. 

### Alternatives

I could not use the plugin and do it an alternative way, e.g. mapSchema()

EmrysMyrddin:
Fixed by #1982

PR #1982, opened 2023-10-04
Support introspection

darren-west:
## Description

Support introspection types in useOnResolve, to be backwards compatible make it an option and disabled by default.

Fixes # [1981](https://github.com/n1ru4l/envelop/issues/1981)

## Type of change

## How Has This Been Tested?

Unit tested

## Checklist:

- [x] I have followed the

### DIFF
--- a/packages/plugins/on-resolve/src/index.ts
+++ b/packages/plugins/on-resolve/src/index.ts
@@ -32,6 +32,10 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
   options: OnResolveOptions<PluginContext>,
 ) => PromiseOrValue<AfterResolver | void>;
 
+export type UseOnResolveOptions = {
+  introspection: Boolean;
+};
+
 /**
  * Wraps the provided schema by hooking into the resolvers of every field.
  *
@@ -39,7 +43,7 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
  */
 export function useOnResolve<PluginContext extends Record<string, any> = {}>(
   onResolve: OnResolve<PluginContext>,
-  excludeIntrospection = true,
+  opts?: UseOnResolveOptions,
 ): Plugin<PluginContext> {
   const hasWrappedResolveSymbol = Symbol('hasWrappedResolve');
   return {
@@ -48,7 +52,7 @@ export function useOnResolve<PluginContext extends Record<string, any> = {}>(
       if (!schema) return; // nothing to do if schema is missing
 
       for (const type of Object.values(schema.getTypeMap())) {
-        if (!(excludeIntrospection && isIntrospectionType(type)) && isObjectType(type)) {
+        if ((opts?.introspection || !isIntrospectionType(type)) && isObjectType(type)) {
           for (const field of Object.values(type.getFields())) {
             if (field[hasWrappedResolveSymbol]) continue;
 
--- a/packages/plugins/on-resolve/src/index.ts
+++ b/packages/plugins/on-resolve/src/index.ts
@@ -33,7 +33,7 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
 ) => PromiseOrValue<AfterResolver | void>;
 
 export type UseOnResolveOptions = {
-  introspection: Boolean;
+  skipIntrospection: boolean;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/plugins/on-resolve/src/index.ts
```diff
@@ -39,6 +39,7 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
  */
 export function useOnResolve<PluginContext extends Record<string, any> = {}>(
   onResolve: OnResolve<PluginContext>,
+  excludeIntrospection = true,
```
EmrysMyrddin: Can you use an option object instead ? If we want to add more options like this in the future, we don't want to force users to give every parameter in the right order.

@ardatan Do you think it is worth a breaking change to replace the `onResolve` parameter with the object option ? Or do you think it's better to place it as a second parameter ?

EmrysMyrddin (changes requested):
Thank you for the PR :-) Very appreciated !

Just made a comment about the proposed API

packages/plugins/on-resolve/src/index.ts
darren-west: Yea no worries I considered it myself @EmrysMyrddin, to me it makes sense to have onResolve as the first argument and options as a second as you must supply an onResolve as otherwise the plugin has no value what do you think?
e.g
```ts
export function useOnResolve<PluginContext extends Record<string, any> = {}>(onResolve: OnResolve, opts?: UseOnResolveOptions): Plugin<PluginContext> {
```
This would keep it backwards compatible, or would you rather I wrap onResolve into the options?

EmrysMyrddin: Both are possible :-) I'm not sure which one is the best. I prefer having one object with everything in it, for consistency with other plugins API, but I'm not sure it is worth the breaking change.

ardatan: I don't think adding an additional parameter with some options object is a breaking change.

EmrysMyrddin: @ardatan Yes, that's the non breaking solution. The question was is it worth to replace the existing parameter with an option object, which would be breaking.

An other solution is to support both passing a function and an object, and you check at runtime in which case we are.

This allows to be non breaking, but still having the same API than other plugins with a unique parameter being an option object.

enisdenjo: > is it worth to replace the existing parameter with an option object, which would be breaking

I don't think we need to do this now, or at all? I think the API of `useOnResolve` is quite good where you provide the onResolve function (which is the goal of the plugin) first and then options optionally second.

@darren-west I'd just change to [this approach](https://github.com/n1ru4l/envelop/pull/1982/files#r1348433831) instead and we're good to go!

enisdenjo (changes requested):

packages/plugins/on-resolve/src/index.ts
darren-west: I will push the changes soon

darren-west: pushed

packages/plugins/on-resolve/src/index.ts
```diff
@@ -32,14 +32,18 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
   options: OnResolveOptions<PluginContext>,
 ) => PromiseOrValue<AfterResolver | void>;
 
+export type UseOnResolveOptions = {
+  introspection: Boolean;
```
enisdenjo: Just "introspection" doesn't say much about what this property does. I'd prefer to name it `skipIntrospection` (and document the default) to make it more clear.

Also, avoid using the `Boolean` alias type (capital B).

```suggestion
  /**
  * Skip executing the `onResolve` hook on introspection queries.
  *
  * @default true
  */
  skipIntrospection: boolean;
```

enisdenjo (changes requested):

packages/plugins/on-resolve/src/index.ts
```diff
@@ -32,13 +32,18 @@ export type OnResolve<PluginContext extends Record<string, any> = {}> = (
   options: OnResolveOptions<PluginContext>,
 ) => PromiseOrValue<AfterResolver | void>;
 
+export type UseOnResolveOptions = {
+  skipIntrospection: boolean;
```
enisdenjo, line 41:
```suggestion
  /**
   * Skip executing the `onResolve` hook on introspection queries.
   *
   * @default true
   */
  skipIntrospection: boolean;
```

enisdenjo (approved):

PR #1986, opened 2023-10-06
Drop support of node 16

EmrysMyrddin:
This drops the support of Node 16 on every packages.
I've also update the CI to run on active LTS (Node 20) and to test against Node 18, 20 and 21.


saihaj (approved):
Should we also update our TS config output target 🤔

EmrysMyrddin:
@saihaj It doesn't seems to work ><' do you know the config I should use ? I'm not an expert of TS config ^^'

ardatan:
I think @saihaj was referring to this one;
https://github.com/n1ru4l/envelop/pull/1986/commits/f113813fd0ae2be04a4ad73a49b7d07efb72c334#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R12

PR #1992, opened 2023-10-09
GraphQL JIT enhancements

ardatan:
- Pin `graphql-jit` version until this is fixed; https://github.com/zalando-incubator/graphql-jit/issues/216
- Remove the default LRU cache **BREAKING**
- Optimize the `enableIf` handling by replacing `async await` with `value-or-promise` check

### DIFF
--- a/packages/plugins/graphql-jit/src/index.ts
+++ b/packages/plugins/graphql-jit/src/index.ts
@@ -1,14 +1,8 @@
 /* eslint-disable no-console */
 import { DocumentNode, ExecutionArgs, ExecutionResult } from 'graphql';
 import { CompiledQuery, compileQuery, CompilerOptions, isCompiledQuery } from 'graphql-jit';
-import {
-  getDocumentString,
-  makeExecute,
-  makeSubscribe,
-  Plugin,
-  PromiseOrValue,
-  TypedExecutionArgs,
-} from '@envelop/core';
+import { ValueOrPromise } from 'value-or-promise';
+import { getDocumentString, makeExecute, makeSubscribe, Plugin } from '@envelop/core';
 
 type JSONStringifier = (result: any) => string;
 
@@ -48,7 +42,7 @@ export const useGraphQlJit = (
 
   const jitCacheByDocument = new WeakMap<DocumentNode, JITCacheEntry>();
 
-  function getCacheEntry<T>(args: TypedExecutionArgs<T>): JITCacheEntry {
+  function getCacheEntry(args: ExecutionArgs): JITCacheEntry {
     let cacheEntry: JITCacheEntry | undefined;
 
     cacheEntry = jitCacheByDocument.get(args.document);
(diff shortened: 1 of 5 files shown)


packages/plugins/graphql-jit/src/index.ts
```diff
@@ -48,10 +44,7 @@ export const useGraphQlJit = (
     cache?: JITCache;
   } = {},
 ): Plugin => {
-  const jitCacheByDocumentString =
-    typeof pluginOptions.cache !== 'undefined'
-      ? pluginOptions.cache
-      : new LRUCache<string, JITCacheEntry>({ max: DEFAULT_MAX, ttl: DEFAULT_TTL });
+  const jitCacheByDocumentString = pluginOptions.cache;
```
EmrysMyrddin, line 41: This means that the cache is now disabled by default ?

packages/plugins/graphql-jit/src/index.ts
```diff
@@ -89,72 +83,77 @@ export const useGraphQlJit = (
...
+      : (cacheEntry.query(
+          args.rootValue,
+          args.contextValue,
+          args.variableValues,
+        ) as PromiseOrValue<ExecutionResultWithSerializer>);
+    if (isPromise(result$)) {
+      return result$.then(r => {
+        r.stringify = cacheEntry.stringify;
+        return r;
+      });
```
EmrysMyrddin: You use `isPromise` instead of `await` to avoid wrapping the result in a Promise if not necessary ?

packages/plugins/graphql-jit/src/index.ts
```diff
@@ -89,72 +83,77 @@ export const useGraphQlJit = (
...
+    }
+    result$.stringify = cacheEntry.stringify;
+    return result$;
```
EmrysMyrddin: Since this pattern happens a lot, perhaps we can make an util function with a name describing what it does ? Because the syntax noise obfuscate a bit the logic i think.

Something like 
```ts
function handlePromiseOrValue<P, R>(p: PromiseOrValue<P>, handler): PromiseOrValue<R> {
	if(isPromise(p) {
		return p.then(value => handler(value))
	}
	return handler(p)
}
```

Which would lead to this usage here:
```suggestion
    return handlePromiseOrValue(result$, result => {
      result.stringify = cacheEntry.stringify
      return result
    })
```

packages/plugins/graphql-jit/src/index.ts
ardatan: Yes, we no longer provide a default LRU cache.
For Yoga, users can rely on the WeakMap based on DocumentNode which is cached by parser cache of Yoga itself.

packages/plugins/graphql-jit/src/index.ts
ardatan: Extra async/await affects the performance somehow in Node.js :/

packages/plugins/graphql-jit/src/index.ts
ardatan: Actually we can use `value-or-promise` package.

PR #2057, opened 2023-11-02
(feat): Add @envelop/response-cache-cloudflare-kv

### LINKED ISSUE #2055: [RFC - New Plugin] Envelop Response Cache Cloudflare KV
## Problem statement
`@envelop/response-cache` already has a great ecosystem, however for those of us building graphql servers on Cloudflare workers, it is invaluable to be able to use the existing tools within the ecosystem. Specifically, Cloudflare KV is an ideal toolset to use for saving response caches as it is a globally distributed low latency KV store.

The goal is to create a package - `@envelop/response-cache-cloudflare-kv` which acts as a companion to `@envelop/response-cache`, allowing users to use Cloudflare KV as a datastore.

## Usage Example
The API exposed by the KV package should closely mirror the existing setup for envelop response cache plugins, e.g the redis cache.
Here is how I envision the KV cache would be setup in a sample worker script
```ts
import { createSchema, createYoga, YogaInitialContext } from 'graphql-yoga';
import { useResponseCache } from '@envelop/response-cache';
import { createKvCache } from 'envelop-response-cache-cloudflare-kv';
import { typeDefs } from './graphql-schema/typeDefs.generated';
import { resolvers } from './graphql-schema/resolvers.generated';

export type Env = {
  GRAPHQL_RESPONSE_CACHE: KVNamespace;
};
export type GraphQLContext = YogaInitialContext & Env & ExecutionContext;

export default {
  async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
    const kvCache = createKvCache({
      KV: env.GRAPHQL_RESPONSE_CACHE,
      ctx,
      keyPrefix: 'graphql', // optional
    });

    const graphqlServer = createYoga<GraphQLContext>({
      schema: createSchema({ typeDefs, resolvers }),
      plugins: [
        useResponseCache({
          cache: kvCache,
          session: () => null,
          includeExtensionMetadata: true,
          ttl: 1000 * 10, // 10 seconds
        }),
      ],
    });

    return graphqlServer.fetch(request, env, ctx);
  },
};

```

## Considerations for development
One key factor to consider here is how to handle saving and invalidating entity keys within the cache.
Let's take an example from the [blog post](https://the-guild.dev/blog/graphql-response-caching-with-envelop) explaining grapqhql caching.

```
Entity   List of Operation cache keys that reference a entity
User:1   OperationCacheKey1, OperationCacheKey2, ...
User:2   OperationCacheKey2, OperationCacheKey3, ...
User:3   OperationCacheKey3, OperationCacheKey1, ...
```

When saving and invalidating, a naïve approach would be to  save and retrieve a list, essentially creating a map of entity key to operation key. E.g
```ts
// during a set() operation
for (const entity of entities) {
  const entityKey = buildEntityKey(entity.typename, entity.id);
  const existingEntityMap = await config.KV.get<string[]>(entityKey, 'json'); // returns an array of operation keys
  if (!entityMap.includes(operationKey)) {
    entityMap.push(operationKey);
  }
  await config.KV.put(entityKey, JSON.stringify(entityMap));
}

// during an invalidate() operation
const entityKey = buildEntityKey(entity.typename, entity.id);
const entityMap = await config.KV.get<string[]>(entityKey, 'json');
if (entityMap) {
  for (const operationId of entityMap) {
    await config.KV.delete(operationId);
  }
}
await config.KV.delete(entityKey);
```

However, this will lead to race conditions, as KV takes up to 60 seconds to propagate changes across Cloudflare's global network.
This means reading and writing to an array stored in a KV key can lead to some inconsistencies, as only the latest write operation will win.

There are two solutions here
1. Use a synchronisation mechanism like [Cloudflare Durable Objects](https://developers.cloudflare.com/durable-objects/)
2. Flatten the entity map into a list of keys, and using `kv.list({prefix: 'keyPrefix'}` to read all operation keys associated with an entity key.

I've taken approach 2, as it requires less setup from a user using this plugin (in option 1 they would have to also provision a durable object).
This means that instead of storing an entity map of `User:1 => [...arrayOfOperationKeys]` we are storing a list of keys that look like this `['User:1:operationKey', 'User:2:operationKey', ...]`

## Making a PR
I already have a basic implementation of this plugin, I will be working over the next couple of weeks to make it production ready and do a lot of testing
I wanted to ask the maintainers / contributors to this project, would you be willing to accept a PR to create the plugin described above in this project.
Happy to connect / have further discussions on anything that is of concern.

Really looking forward to hearing more. Loved working with envelop and yoga so far, keen to give something back to the ecosystem 😄

AdiRishi:
## Description

This PR creates a new package `@envelop/response-cache-cloudflare-kv` which acts as a companion to @envelop/response-cache, allowing users to use Cloudflare KV as a datastore.
This is ideal for users making graphql servers in Cloudflare workers since it gives them a globally distributed low latency data store.

Fixes #2055 

## Type of change

## How Has This Been Tested?
This section is a work in progress
So far I have tested this plugin externally with detailed vitest unit tests. I'm trying to work out how to do the necessary setup for unit tests in this environment.

## Checklist:

- [x] I have followed the

## Further comments
I've left a detailed description of the proposed change and some engineering discussions in the linked issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "prepare": "husky install",
     "prettier": "prettier --cache --ignore-path .prettierignore --write --list-different .",
     "release": "pnpm build && changeset publish",
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
     "test:ci": "jest --coverage",
     "test:core": "jest ./packages/core --coverage",
     "ts:check": "tsc --noEmit"
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -1,9 +1,9 @@
 import type { ExecutionResult } from 'graphql';
 import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types';
 import type { Cache, CacheEntityRecord } from '@envelop/response-cache';
-import { buildOperationKey } from './cache-key';
-import { invalidate } from './invalidate';
-import { set } from './set';
+import { buildOperationKey } from './cache-key.js';
+import { invalidate } from './invalidate.js';
+import { set } from './set.js';
 
 export type KvCacheConfig = {
   /**
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -36,12 +36,19 @@ export type KvCacheConfig = {
  * @returns A cache object that can be passed to envelop's `useResponseCache` plugin
  */
 export function createKvCache(config: KvCacheConfig): Cache {
+  if (config.cacheReadTTL && config.cacheReadTTL < 60000) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'Cloudflare KV cacheReadTTL must be at least 60000 (60 seconds). Using default value of 60000 instead.',
+    );
+  }
+  const computedTtlInSeconds = Math.max(Math.floor((config.cacheReadTTL ?? 60000) / 1000), 60); // KV TTL must be at least 60 seconds
+
   const cache: Cache = {
     async get(id: string) {
-      const ttlInSeconds = Math.max(Math.floor((config.cacheReadTTL ?? 60000) / 1000), 60); // KV TTL must be at least 60 seconds
       const kvResponse = await config.KV.get(buildOperationKey(id, config.keyPrefix), {
         type: 'text',
-        cacheTtl: ttlInSeconds,
+        cacheTtl: computedTtlInSeconds,
       });
       if (kvResponse) {
         return JSON.parse(kvResponse) as ExecutionResult;
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -1,0 +1,72 @@
+import type { ExecutionResult } from 'graphql';
+import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types';
+import type { Cache, CacheEntityRecord } from '@envelop/response-cache';
+import { buildOperationKey } from './cache-key';
+import { invalidate } from './invalidate';
+import { set } from './set';
+
+export type KvCacheConfig = {
+  /**
+   * The Cloudflare KV namespace that should be used to store the cache
+   */
+  KV: KVNamespace;
+  /**
+   * The Cloudflare worker execution context. Used to perform non-blocking actions like cache storage and invalidation.
+   */
+  ctx: ExecutionContext;
+  /**
+   *  Defines the length of time in milliseconds that a KV result is cached in the global network location it is accessed from.
+   *
+   * The cacheTTL parameter must be an integer greater than or equal to 60000 (60 seconds), which is the default.
+   */
+  cacheReadTTL?: number;
+  /**
+   * A prefix that should be added to all cache keys
+   */
+  keyPrefix?: string;
+};
+
+/**
+ * Creates a cache object that uses Cloudflare KV to store GraphQL responses.
+ * This cache is optimized for Cloudflare workers and uses the `ctx.waitUntil` method to perform non-blocking actions where possible
+ *
+ * To find out more about how this cache is implemented see https://the-guild.dev/blog/graphql-response-caching-with-envelop
+ *
+ * @param config Modify the behavior of the cache as it pertains to Cloudflare KV
+ * @returns A cache object that can be passed to envelop's `useResponseCache` plugin
+ */
+export function createKvCache(config: KvCacheConfig): Cache {
+  const cache: Cache = {
+    async get(id: string) {
+      const ttlInSeconds = Math.max(Math.floor((config.cacheReadTTL ?? 60000) / 1000), 60); // KV TTL must be at least 60 seconds
+      const kvResponse = await config.KV.get(buildOperationKey(id, config.keyPrefix), {
+        type: 'text',
+        cacheTtl: ttlInSeconds,
+      });
+      if (kvResponse) {
+        return JSON.parse(kvResponse) as ExecutionResult;
+      }
+      return undefined;
+    },
+
+    set(
+      /** id/hash of the operation */
+      id: string,
+      /** the result that should be cached */
+      data: ExecutionResult,
+      /** array of entity records that were collected during execution */
+      entities: Iterable<CacheEntityRecord>,
+      /** how long the operation should be cached (in milliseconds) */
+      ttl: number,
+    ) {
+      // Do not block execution of the worker while caching the result
+      config.ctx.waitUntil(set(id, data, entities, ttl, config));
+    },
+
+    invalidate(entities: Iterable<CacheEntityRecord>) {
+      // Do not block execution of the worker while invalidating the cache
+      config.ctx.waitUntil(invalidate(entities, config));
+    },
+  };
+  return cache;
+}
--- a/packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
@@ -11,7 +11,7 @@ export async function invalidate(
   for (const entity of entities) {
     const entityKey = buildEntityKey(entity.typename, entity.id, config.keyPrefix);
 
-    for await (const kvKey of _getAllKvKeysForPrefix(entityKey, config)) {
+    for await (const kvKey of getAllKvKeysForPrefix(entityKey, config)) {
       if (kvKey.metadata?.operationKey) {
         kvPromises.push(config.KV.delete(kvKey.metadata?.operationKey));
         kvPromises.push(config.KV.delete(kvKey.name));
@@ -22,7 +22,7 @@ export async function invalidate(
   await Promise.allSettled(kvPromises);
 }
 
-export async function* _getAllKvKeysForPrefix(prefix: string, config: KvCacheConfig) {
+export async function* getAllKvKeysForPrefix(prefix: string, config: KvCacheConfig) {
   let keyListComplete = false;
   let cursor: string | undefined;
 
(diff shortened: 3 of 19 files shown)


AdiRishi:
I'm having a little bit of trouble with the unit tests. As I was developing this package externally, I already have a great set of unit tests written in vitest.

While converting to Jest, I noticed I'd have to use a different jest environment - https://miniflare.dev/testing/jest
But it looks like the entire project uses the main jest.config.js file.
Any tips on how you would like this setup?
If you would rather I not use jest, I can just have that local folder use vitest and write the tests that way.
Here is the package + tests setup in another project if you're curious on how the tests are to be written - https://github.com/AdiRishi/tiptop/tree/main/packages/envelop-response-cache-cloudflare-kv (where I originally developed this)

Keen for comments :)

AdiRishi:
Update: I think I've figured out how to do testing with the setup present in this repo. I'll update the PR soon

AdiRishi:
Question, should I add a changesets release here or in a new PR?

package.json
```diff
@@ -19,7 +19,7 @@
     "prepare": "husky install",
     "prettier": "prettier --cache --ignore-path .prettierignore --write --list-different .",
     "release": "pnpm build && changeset publish",
-    "test": "jest",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest",
```
AdiRishi, line 22: For anyone reviewing, this is needed for the miniflare test environment - https://miniflare.dev/testing/jest
It doesn't cause any breaking changes in this project

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "incremental": true,
     "baseUrl": ".",
     "outDir": "dist",
-    "rootDir": "packages",
+    "rootDir": ".",
```
AdiRishi: This was needed because Jest complained about not having access to `@cloudflare/workers-types` which exists in `node_modules` in the actual root directory. This is likely because the built version of this package has .ts files which jest thinks it needs to compile.
This change has no impact on existing type checks / other functionality in this repo. Tests + lint + ts:check all pass

package.json
EmrysMyrddin: @enisdenjo Isn't this that was causing memory leaks ?

packages/plugins/response-cache-cloudflare-kv/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * To find out more about how this cache is implemented see https://the-guild.dev/blog/graphql-response-caching-with-envelop
+ *
+ * @param config Modify the behavior of the cache as it pertains to Cloudflare KV
+ * @returns A cache object that can be passed to envelop's `useResponseCache` plugin
+ */
+export function createKvCache(config: KvCacheConfig): Cache {
+  const cache: Cache = {
+    async get(id: string) {
+      const ttlInSeconds = Math.max(Math.floor((config.cacheReadTTL ?? 60000) / 1000), 60); // KV TTL must be at least 60 seconds
```
EmrysMyrddin: You probably want to move this outside of the get function, in the `creatKvCache` body directly to avoid computing it each time we read the cache ?

And perhaps some kind of warning would be great to let the user know their configuration is probably not working as they expect :-)

AdiRishi:
Alright, I can see the ESM check is failing, this seems to be because `pnpm build` doesn't actually output any built files for all packages.
It looks like the change to the root `tsconfig.json` `rootDir` caused that, but I'm not sure why, since I don't completely get how `bob` is working. I'll try and debug.

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (!kvListResponse.list_complete) {
+      cursor = kvListResponse.cursor;
+    }
+
+    for (const keyResult of kvListResponse.keys) {
+      yield keyResult;
+    }
+  } while (!keyListComplete);
+}
```
EmrysMyrddin, line 54: Perhaps you can use a for loop here ?

```suggestion
  for(let kvListResponse: ResponseType | undefined ; kvListResponse?.list_complete;) {
    const kvListResponse = await config.KV.list<{ operationKey: string }>({
      prefix,
      cursor: kvListResponse.cursor,
    });
    
    for (const keyResult of kvListResponse.keys) {
      yield keyResult;
    }
  }
}
```

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  for (const entity of entities) {
+    const entityKey = buildEntityKey(entity.typename, entity.id, config.keyPrefix);
+
+    for await (const kvKey of _getAllKvKeysForPrefix(entityKey, config)) {
+      if (kvKey.metadata?.operationKey) {
+        kvPromises.push(config.KV.delete(kvKey.metadata?.operationKey));
+        kvPromises.push(config.KV.delete(kvKey.name));
+      }
+    }
```
EmrysMyrddin: I like this trick 😃

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  await Promise.allSettled(kvPromises);
```
EmrysMyrddin: Perhaps we could even do this in an async function to allow parallelisation ?

```suggestion
  await Promise.allSettled(entities.map(async entity => {
    const kvPromises: Promise<unknown>[] = []; // Collecting all the KV operations so we can await them all at once
    const entityKey = buildEntityKey(entity.typename, entity.id, config.keyPrefix);

    for await (const kvKey of _getAllKvKeysForPrefix(entityKey, config)) {
      if (kvKey.metadata?.operationKey) {
        kvPromises.push(config.KV.delete(kvKey.metadata?.operationKey));
        kvPromises.push(config.KV.delete(kvKey.name));
      }
    }
    
    return Promise.allSettled(kvPromises);
  }));
```

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export async function* _getAllKvKeysForPrefix(prefix: string, config: KvCacheConfig) {
```
EmrysMyrddin: We generally don't use `_` prefix for functions. A function should be either considered private and not exported or public and exported. I think it is ok to have it exported without the `_` prefix :-)

AdiRishi:
> Alright, I can see the ESM check is failing, this seems to be because `pnpm build` doesn't actually output any built files for all packages. It looks like the change to the root `tsconfig.json` `rootDir` caused that, but I'm not sure why, since I don't completely get how `bob` is working. I'll try and debug.

Ok, `bob build` is outputting files now. But the ESM output for the package I added is still failing. Seems like the generated `esm` files don't import others with a `.js` extension at the end. Looking into it.

EmrysMyrddin (commented):
That's really great ! Very appreciated :-)

I just leaved some minor comments

package.json
enisdenjo: it causes Jest's LeakDetector to misbehave. since we're not using `--detectLeaks` here, we should be ok

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
AdiRishi: I personally find the `do ... while` loop more readable. But I'm happy to switch this if you feel strongly about it.

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
AdiRishi: That's a great catch!

packages/plugins/response-cache-cloudflare-kv/src/invalidate.ts
EmrysMyrddin: If you think it's more readable, let it that way :-)

package.json
EmrysMyrddin: Thank you for the insight :-)

AdiRishi:
Nice, looks like the esm bug is fixed, all checks pass and PR comments resolved. Thank you for the super helpful code review, caught quite a few things I totally missed 😄 

I'm not really sure how to go from this PR -> deploying the package to NPM.

Once the package is deployed, I'll test it myself for a bit in production before making a PR to bump it to 1.0

EmrysMyrddin:
Great Work ! 

I'm not sure it is possible to release alpha version based on a fork branch. Let me check if I can do something for this :-)

EmrysMyrddin:
You can try out `@envelop/response-cache-cloudflare-kv@0.2.0-alpha-20231106140646-27776100` :-)

AdiRishi:
@EmrysMyrddin thanks for the alpha deployment!

I've done some testing, found some minor issues / bugs
- Due to some typescript weirdness, when you pass in your argument for `KV` which is of type `KVNamespace` typescript rejects this. I think this is because normal workers projects don't import the type `KVNamespace` but rather have it implicitly defined in the tsconfig.
- I can't tell if this is a bug in my library or envelop response cache but it DOES NOT make a unique cache per session when the `session()` factory is defined. No matter the value given in session it looks at the same cache.

However, the basic functionality in the package works as expected, it reads, writes and invalidates cache successfully.

Given that, I think let's go ahead and close this PR but release the package as a minor version. Maybe `0.2.0` but without the alpha tag?
I'll keep investigating and open new issues / PRs for the bugs I discover.

AdiRishi:
Rebased this PR off the latest commits made to `main`. There were some conflicts in `pnpm-lock.yaml` so it had to be done

AdiRishi:
Hey all, just wanted to ask for an update. Is there a plan to merge this soon?

EmrysMyrddin:
Hey ! Really sorry to have dropped the ball here.

Yeah, let's merge this as an "beta" version and investigate the issues you pointed out.

EmrysMyrddin (approved):

EmrysMyrddin:
Hey @AdiRishi, Did you have the opportunity to investigate further the last issues missing for a full release ?

AdiRishi:
Hey @EmrysMyrddin , so sorry I dropped the ball here!

I think I have some solutions to the problems, I'll make some PRs soon.

Edit:

For the typing issue, I have a theory, I think the workers-types package needs to be a peer-dependency. I did a similar thing in another package I built and it seemed to work...so I'll try again with a PR.

EmrysMyrddin:
Great to hear! You commitment to this feature is very appreciated!

PR #2090, opened 2023-11-22
fix: [`@envelop/response-cache`] `ttl` yielded as `NaN`

### LINKED ISSUE #2087: [`@envelop/response-cache`] `ttl` yielded as `NaN`
### Issue workflow progress

- [x] 1. The issue provides a
     [Stackblitz](https://stackblitz.com/edit/node-afnuvb?file=index.js).
---

**Describe the bug**

The ttl that envelop response cache sends to REDIS is `NaN` which result in never expiring keys.

1. Go to stackblitz
2. node index.js
3. Open graphiql
4. Run the query
5. Look in console, it will say
```
CACHE SET b62dfad0b49bda5568f9c87772c29850df5b20ded8edf8c075e27c75658757da NaN
Error: ttl is NaN
    at Object.set (/home/projects/node-z8heaa/index.js:86:15)
    at maybeCacheResult (/home/projects/node-z8heaa/node_modules/@envelop/response-cache/esm/plugin.js:294:23)
    at onExecuteDone (/home/projects/node-z8heaa/node_modules/@envelop/response-cache/esm/plugin.js:304:28)
    at eval (/home/projects/node-z8heaa/node_modules/@envelop/core/esm/orchestrator.js:406:42)
    at async YogaServer.getResultForParams (/home/projects/node-z8heaa/node_modules/graphql-yoga/esm/server.js:283:26)
    at async handle (/home/projects/node-z8heaa/node_modules/graphql-yoga/esm/server.js:338:25)
    at async handlerWithErrorHandling (/home/projects/node-z8heaa/node_modules/@whatwg-node/server/esm/plugins/useErrorHandling.js:45:38)
```

**Expected behavior**

I expect ttl to always be a valid number and never NaN. In this case probably 0. Or maybe this function shouldn't even be invoked since ttl === 0 iirc when I look at source code means that we do not set the cache entry at all.

**Environment:**

- OS: Mac
- NodeJS: 18.18
- "@graphql-yoga/plugin-response-cache": "3.1.0",
- "graphql-yoga": "5.0.0"

**Additional context**

This was originally reported in a comment in related issue https://github.com/n1ru4l/envelop/issues/1968#issuecomment-1746433561

Relevant code:
```ts
            function maybeCacheResult(result, setResult) {
                processResult(result.data);
                // we only use the global ttl if no currentTtl has been determined.
                const finalTtl = currentTtl ?? globalTtl; // <--- currentTttl is `NaN`, globalTtl is 0
                if (skip || !shouldCacheResult({ cacheKey, result }) || finalTtl === 0) {
                    if (includeExtensionMetadata) {
                        setResult(resultWithMetadata(result, { hit: false, didCache: false }));
                    }
                    return;
                }
                
                cache.set(cacheKey, result, identifier.values(), finalTtl);
                if (includeExtensionMetadata) {
                    setResult(resultWithMetadata(result, { hit: false, didCache: true, ttl: finalTtl }));
                }
            }
```

klippx:
## Description

Ensures that response cache is not caching with a TTL of `NaN`

Fixes #2087 

## Type of change

Adding links to sandbox or providing screenshots can help us understand more about this PR and take
action on it as appropriate

## How Has This Been Tested?

- [x] Reproduced issue of getting TTL in a spec and fixed it
- [x] Added an extra test for Infinity

## Checklist:

- [x] I have followed the
~~- [ ] I have commented on my code, particularly in hard-to-understand areas~~
~~- [ ] I have made corresponding changes to the documentation~~
~~- [ ] New and existing unit tests pass locally with my changes~~
~~- [ ] Any dependent changes have been merged and published in downstream modules~~


ardatan:
Can we add a failing test before discussing the solution? Because it looks something might be wrong internally so it gives NaN in somewhere? Also I see some changes that removes console.warn which is unrelated to the context of this PR, right?

klippx:
> Can we add a failing test before discussing the solution? Because it looks something might be wrong internally so it gives NaN in somewhere? 

I did add a breaking test that teases out the `NaN`, that is clear in the commit history: https://github.com/n1ru4l/envelop/pull/2090/commits/03e839fbc877e737c5a62c598971abf1717e67c5

I made this commit history for your benefit, to give context of my problem and so you could understand the issue and understand the fix. I have also prior to this opened and referenced the relevant issue for it, but now you make it sound like I just opened a random PR with a random change 😢

There are many issues here such as this `const ttl = cacheControl.maxAge * 1000;` you are not considering that this `cacheControl.maxAge ` might be undefined, typescript is lying to you and saying this is a number because of bad typings:
- `const cacheControlAnnotations: Record<string, any>[] | undefined`
- => `cacheControl.maxAge: any`
- => cannot safely treat this as a number and do `* 1000`.

Further issues:
- `currentTtl = calculateTtl(ttlPerType[typename], currentTtl);`
  - This is dangerously casting the result of an unchecked index access to a function that expects a guaranteed `number` when this 100% can be `undefined`
  - In fact, this is more often that not gonna get a `undefined` here so this totally throws a spanner in the works of `calculateTtl`

Do you need me to add breaking tests for this as well or is common sense enough?

> Also I see some changes that removes console.warn which is unrelated to the context of this PR, right?

Yes, I can remove that.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -331,7 +331,8 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           [MapperKind.COMPOSITE_TYPE]: type => {
             const cacheControlAnnotations = getDirective(schema, type, 'cacheControl');
             cacheControlAnnotations?.forEach(cacheControl => {
-              const ttl = cacheControl.maxAge * 1000;
+              const ttl = cacheControl.maxAge != null ? cacheControl.maxAge * 1000 : null;
+
               if (ttl != null) {
                 ttlPerType[type.name] = ttl;
               }
```
EmrysMyrddin: Do we really need the local variable here ?
```suggestion
              if (cacheControl.maxAge != null) {
                ttlPerType[type.name] = cacheControl.maxAge * 1000;
              }
```

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -354,7 +355,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           if (cacheControlDirective) {
             const cacheControlAnnotations = getDirective(schema, fieldConfig, 'cacheControl');
             cacheControlAnnotations?.forEach(cacheControl => {
-              const ttl = cacheControl.maxAge * 1000;
+              const ttl = cacheControl.maxAge != null ? cacheControl.maxAge * 1000 : null;
               if (ttl != null) {
                 ttlPerSchemaCoordinate[schemaCoordinates] = ttl;
               }
```
EmrysMyrddin: Same here, we probably don't need the local variable :-)
```suggestion
              if (cacheControl.maxAge != null) {
                ttlPerSchemaCoordinate[schemaCoordinates] = cacheControl.maxAge * 1000;
              }
```

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -410,7 +411,10 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
 
           types.add(typename);
           if (typename in ttlPerType) {
-            currentTtl = calculateTtl(ttlPerType[typename], currentTtl);
+            const maybeTtl = ttlPerType[typename] as number | undefined;
+            if (maybeTtl != null && Number.isNaN(maybeTtl) === false) {
+              currentTtl = calculateTtl(maybeTtl, currentTtl);
+            }
```
EmrysMyrddin: To avoid duplication, perhaps`calculateTtl` can do this check by taking either `unknow` or `number | undefined | null` as a first parameter ?

EmrysMyrddin:
The issue here is that the code doesn't handle properly all the use cases of `@cacheControl` directive. This was already fixed in a previous version, but due to a lack of type checking the bug came back.

Do you think you can do something about it ?

I'm thinking of two things we can do:
 - in the `onSchemaChange` hook, check that the definition of the `@cacheControl` is correct. This would at least allow us to avoid checking everywhere the type of the arguments and safely cast them.
 - create a `getCacheControlDirective` which would return a well typed object like `{ maxAge?: number, scope?: string }`, relying on the fact that we already checked that the directive is well defined

This would improve the maintainability of this feature :-) But if you think you don't have time for it, we can create an issue to do it in an other PR.

ardatan:
@klippx 
I (and everyone else here I am 100% sure) appreciate so much for your collaboration and contribution. NEVER EVER any contribution is "random" but so valuable for us. I apologize if my comment there made you feel in this way. That's entirely my mistake. 
I am also sorry I missed your changes with the test.

packages/plugins/response-cache/src/plugin.ts
klippx: Fixed, please mark as resolved if you are happy.

packages/plugins/response-cache/src/plugin.ts
klippx: Fixed, please mark as resolved if you are happy.

packages/plugins/response-cache/src/plugin.ts
klippx: Fixed, pushed a new commit. Please mark as resolved if it looks good.

klippx:
@EmrysMyrddin 

> The issue here is that the code doesn't handle properly all the use cases of `@cacheControl` directive. This was already fixed in a previous version, but due to a lack of type checking the bug came back.
> 
> Do you think you can do something about it ?

Thanks for the great review, and the encouragement to take on the challenge of making it even better.

Regarding validation:
- I think it was too hard for me to figure out how to do the nuts and bolts of how the inner working of actually validating the directive correctly so it falls a bit short in that regard, I decided to not try since I was getting myself into a quite deep hole.
- With some more tutoring or hand holding I can still make a second shot. For me I had some confusion about the outer `schema.getDirective` and the inner function `getDirective` and where validation can/should be done (I tried in the outer `schema.getDirective` scope)
- TLDR: I tried, but failed. But can try again with help.

Regarding returning a well typed object:
- I pushed a commit that essentially is type casting (you can argue unsafely, since we do not have the validation component mentioned above) to what the expected type would be. 
- This is an alternative to the `any` of today, and in my mind it is at least is not WORSE and type wise (as long as we are not obscuring or being misleading in the types here) it might better express the expectations of the data and intent of the following code consuming it.
- [x] Suggestion: I can delete the `getCacheControlDirective` and simply inline it instead.

I also renamed some `cacheControlDirective` local variables to not accidentally shadow the exported `cacheControlDirective`, I can revert that if you want.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -172,7 +178,7 @@ export const defaultBuildResponseCacheKey = (params: {
 export const defaultShouldCacheResult: ShouldCacheResultFunction = (params): Boolean => {
   if (params.result.errors) {
     // eslint-disable-next-line no-console
-    console.warn('[useResponseCache] Failed to cache due to errors');
+    // console.warn('[useResponseCache] Failed to cache due to errors');
```
klippx: I keep accidentally pushing this, I personally hate it I guess

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -351,12 +363,15 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
           }
 
-          if (cacheControlDirective) {
-            const cacheControlAnnotations = getDirective(schema, fieldConfig, 'cacheControl');
+          if (directive) {
+            const cacheControlAnnotations = getDirective(
+              schema,
+              fieldConfig,
+              'cacheControl',
+            ) as unknown as CacheControlDirective[] | undefined;
```
EmrysMyrddin, line 372: The idea of having a `getCacheControlDirective` was to hide the noise of the type casting from the actual which is using it.

I think it could make the code clearer and avoid the duplication since this is done twice :-)

klippx: On the flip side, then we have an extra function which is an abstractional cost, and all it does is type casting. I believe that this is in a gray zone but imo the layer of indirection and unnecessary jump / context switch while reading code does not justify us trying to keep things too "DRY".

EmrysMyrddin: Ok then let's keep it like this :-)

EmrysMyrddin:
There is a little conflict to resolve and let's merge this.
Sorry for the time it took 😞

EmrysMyrddin:
Thank you again for this :-)

PR #2093, opened 2023-11-23
fix(response-cache): run enabled only in execution

### LINKED ISSUE #2034: Can no longer use result of `useGenericAuth` to determine if caching should be enabled in response cache plugin
### Issue workflow progress

- [ ] 1. The issue provides a
     [Stackblitz](https://stackblitz.com/fork/node).
- [ ] 2. A failing test has been provided
- [ ] 3. A local solution has been provided
- [ ] 4. A pull request is pending review

---

**Describe the bug**

Previously, a configuration for envelop plugins like the below worked fine, `currentUser` was made available by the `useGenericAuth` plugin at runtime.

```ts
      useGenericAuth({
        contextFieldName: 'currentUser',
        resolveUserFn,
        mode: 'resolve-only',
      }),
      useResponseCache({
        // Disable caching for `Admin` tokens hitting the attendee API (should
        // never happen) and for `ImpersonatedAttendee` tokens.
        // `ImpersonatedAttendee` tokens are intended to be used when
        // previewing content during the authoring experience and so ought to
        // always see the latest view of data.
        enabled: (context: Context) =>
          context.currentUser.role !== 'Admin' &&
          context.currentUser.role !== 'ImpersonatedAttendee',
        // Setting 2000 basically to debounce requests.
        ttl: 2000,
        session: () => null,
      }),
```

The check for whether the `useResponseCache` plugin is enabled in the `onParse` lifecycle function as part of https://github.com/n1ru4l/envelop/pull/1977/commits/b4e1c117e373adcd43fede05475f5387cbfd3db2 means `currentUser` is no longer available in the `enable` function of the `useResponseCache` plugin.

1. Configure envelop with `@envelop/generic-auth`, and `@envelop/response-cache`
2. Attempt to access the user from the generic-auth plugin in the `enabled` function of the response cache plugin
3. Make any call to the graphql endpoint
4. Observe every result is a 400 response with no error information

Minimal reproduction can be found at https://stackblitz.com/edit/node-3zuq7n

**Expected behavior**

The `currentUser` from `useGenericAuth` should be available in the `enabled` function when it is executed.

**Environment:**

- OS: macOS 14
- NodeJS: v18.17.0
- `@envelop/*` versions:
  -    "@envelop/core": "^5.0.0",
  -    "@envelop/generic-auth": "^7.0.0",
  -    "@envelop/on-resolve": "^4.1.0",
  -    "@envelop/response-cache": "^6.0.0",

**Additional context**

The release notes when new versions of envelop plugins were cut on the GitHub releases page did not include anything about the change for `enabled` in the response cache plugin. Even the change set notes indicating a breaking change in https://github.com/n1ru4l/envelop/pull/1977 do not indicate a change in how parameters are used in the plugin lifecycle.

ardatan:
After this change, Response Cache plugin should come after all other plugins modifying the execute fn.
Fixes https://github.com/n1ru4l/envelop/issues/2034

### DIFF
--- a/.changeset/breezy-pots-lick.md
+++ b/.changeset/breezy-pots-lick.md
@@ -1,0 +1,5 @@
+---
+'@envelop/response-cache': patch
+---
+
+Run enabled once only in execution
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -22,7 +22,6 @@ import {
   getDirective,
   MapperKind,
   mapSchema,
-  memoize1,
   memoize4,
   mergeIncrementalResult,
 } from '@graphql-tools/utils';
@@ -205,7 +204,6 @@ export type ResponseCacheExecutionResult = ExecutionResult<
   { responseCache?: ResponseCacheExtensions }
 >;
 
-const originalDocumentMap = new WeakMap<DocumentNode, DocumentNode>();
 const addEntityInfosToDocument = memoize4(function addTypeNameToDocument(
   document: DocumentNode,
   addTypeNameToDocumentOpts: { invalidateViaMutation: boolean },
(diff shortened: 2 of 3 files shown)


EmrysMyrddin:
Closed in favor of #2094

PR #2157, opened 2024-02-19
add `usePrunedSchema` plugin

EmrysMyrddin:
## Description

Adds a plugin to automatically clean up the schema by applying [`pruneSchema` from `@graphql-tools/utils`](https://the-guild.dev/graphql/tools/docs/api/modules/utils_src#pruneschema).

## Type of change

### DIFF
--- a/packages/plugins/pruned-schema/README.md
+++ b/packages/plugins/pruned-schema/README.md
@@ -1,0 +1,34 @@
+## `@envelop/pruned-schema`
+
+This plugins clean up the registered schema by removing unused and empty types.
+
+## Getting Started
+
+```bash
+yarn add @envelop/pruned-schema
+```
+
+## Usage Example
+
+See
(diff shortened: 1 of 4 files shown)


saihaj (commented):
why would we want to do this on runtime, why not have a pruned schema at build time before the server even starts? This way the pruned schema is also part of the schema registry?

EmrysMyrddin:
This is probably for when you are not fully in control of the schema you are exposing, like federating multiple schemas from other teams in your company :-)

But If you think this is not useful I can close it. I'm not that much convinced of the utility of this, it should perhaps better be applied on schema loading. But this has the good aspect of being very simple to add to an existing setup. Just put the plugin and the schema is cleaned up.

saihaj (approved):
I see what you mean. I see how it can be easy to integrate

PR #2201, opened 2024-04-01
Improvements for opentelemetry plugin

dotansimha:


packages/plugins/opentelemetry/README.md
```diff
@@ -35,8 +35,38 @@ const getEnveloped = envelop({
...
+```ts
+import { execute, parse, specifiedRules, subscribe, validate } from 'graphql'
+import { envelop, useEngine } from '@envelop/core'
+import { useOpenTelemetry } from '@envelop/opentelemetry'
+import { trace } from '@opentelemetry/api'
+import { BasicTracerProvider, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base'
+
+const getEnveloped = envelop({
+  plugins: [
+    useEngine({ parse, validate, specifiedRules, execute, subscribe }),
```
EmrysMyrddin, line 55: Not sure you have to put the `useEngine` here, it can be part of the "... other plugins ...".
```suggestion
```

EmrysMyrddin: I just realized that you probably want to have a copy-pastable working example...
Then it's probably ok the leave it ^^

packages/plugins/opentelemetry/README.md
```diff
@@ -35,8 +35,38 @@ const getEnveloped = envelop({
...
+        variables: true, // Includes the operation variables values as part of the metadata collected
+        result: true // Includes execution result object as part of the metadata collected
+      },
+      trace.getTracerProvider()
+    )
+  ]
+})
+```
+
+This example integrates Jaeger tracer:
```
EmrysMyrddin: We should probably have a working example in the repo we can link to. It would also allow us to write actual integration tests for it.

dotansimha: Yeah, sure, on it! 
I think we can also just have a simple command to run Jaeger. Let me add that.

dotansimha: done!

packages/plugins/opentelemetry/src/index.ts
```diff
@@ -131,6 +135,17 @@ export const useOpenTelemetry = (
...
+
+      const shouldReport = options.excludedOperationNames
+        ? typeof options.excludedOperationNames === 'function'
+          ? !options.excludedOperationNames(operationAst.name?.value)
+          : !options.excludedOperationNames.includes(operationAst.name?.value || '')
+        : true;
+
+      if (!shouldReport) {
+        return;
+      }
```
EmrysMyrddin: Having tow ternary like this is a bit difficult to read. Since the result is just used to return, I do think it's a bit easier like this, what do you think ?

```suggestion
      if (options.excludedOperationNames) {
        if (typeof options.excludedOperationNames === 'function')
          if (!options.excludedOperationNames(operationAst.name?.value)) {
            return
          }
        } else if (!options.excludedOperationNames.includes(operationAst.name?.value || '')) {
          return
        }
      }
```

or even something like this, but a bit weird to have a ternary inside a if condition ^^'
```suggestion
      if (options.excludedOperationNames) {
        if (typeof options.excludedOperationNames === 'function'
          ? !options.excludedOperationNames(operationAst.name?.value)
          : !options.excludedOperationNames.includes(operationAst.name?.value || '')
        ) {
          return
        }
      }

EmrysMyrddin (approved):

packages/plugins/opentelemetry/src/index.ts
dotansimha: i picked the in-between solution :D

Issue #2181, opened 2024-03-19
[useResponseCache] Add the computed `scope` in `buildResponseCacheKey` function params


k-i-k-s:
### The problem 

Currently the scope `PUBLIC` creates a cache key per user, since I include the `session` in the key string.

But I'd like to have only ONE key globally shared when it is `PUBLIC`, and a key per session when it is `PRIVATE`.

### A viable solution

To be able to generate a key that is shared across all users, I'd need a property that tells the final computed scope, named `scope` or `isPrivate`.

Maybe I'm missing something, I didn't find any way to reach this goal with the current state of the plugin.
Is it something you considered or conversely that you don't want to include in the `response-cache` ?

To me it appears like a light solution, that could bring more flexibility when building the cache key.

_This update should be reflected in `@graphql-yoga/plugin-response-cache` too_.

EmrysMyrddin:
If I understand your problem properly, you have a custom key builder right ? And it includes the session id in the key ?

The problem is that with current implementation, giving access to the scope is probably not that easy.

The key is built very early in the pipeline, which prevents us from giving access to some things that are computed later, such as the TTL or the scope.

I have to read the code again to verify if it is doable.

k-i-k-s:
Yes we have to build a custom key because of some context variables needed to fluctuate the key, the session id is one of them.

I checked the code and the key is indeed built early, the scope is computed later from the `result.data`.
It is still possible to get it I think considering that `scopePerSchemaCoordinate` already has the scopes from the schema. 

The final scope could be calculated from parsing the query itself before getting the cache key, then getting the object keys in an array and passing them to the `isPrivate` function. Maybe memoize it to prevent later calculation with the same query.

k-i-k-s:
On `apollo-response-cache` plugin when the scope is `PUBLIC` it saves only one key in cache, the parsing behaviour could replicated from their source code.

EmrysMyrddin:
I have checked, I confirm that this is not possible without sacrificing performance.
Our implementation of response-caching is based on the actual response returned after the execution, which means that the cache key can't be based on the scope.

This allows us for example to entirely skip the parse/validation phases, which can be a major performance boost on medium to large requests.

On Yoga, it also allows us to take advantage of the HTTP caching mechanism, meaning we don't even have to actually consume the body of the request to send back the response on cache hit.

The `scope` is here only to enforce the fact that a response should never be cached if there is no `sessionId` with a `PRIVATE` scope.

[...]

PR #2202, opened 2024-04-02
feat(response-cache): add extras property to get scope and its metadata from buildResponseCacheKey

k-i-k-s:
Issue related : https://github.com/n1ru4l/envelop/issues/2181

## Description

Include a new `extras` property to get `scope` (as a `Proxy` to avoid performance impact when unused) in `buildResponseCacheKey`, allowing more granularity over cache control.

## Type of change

## How Has This Been Tested?

I added some unit tests to check that `PRIVATE` scope is correctly retrieved from a query, when directive `@cacheControl(scope: PRIVATE)` is put on query, field and subfield.

## Checklist:

- [x] I have followed the

### DIFF
--- a/packages/plugins/response-cache/src/get-scope.ts
+++ b/packages/plugins/response-cache/src/get-scope.ts
@@ -10,7 +10,7 @@ import {
   SelectionNode,
   visit,
 } from 'graphql';
-import { memoize1 } from '@graphql-tools/utils';
+import { LRUCache } from 'lru-cache';
 import { isPrivate, type CacheControlDirective } from './plugin';
 
 /** Parse the selected query fields */
@@ -92,14 +92,33 @@ function getSchemaCoordinatesFromQuery(schema: GraphQLSchema, query: string): Se
   return fields;
 }
 
+export type Scope = {
+  scope: NonNullable<CacheControlDirective['scope']>;
+  metadata: { privateProperty?: string; hitCache?: boolean };
+};
+
+const scopeCachePerSchema = new WeakMap<GraphQLSchema, LRUCache<string, Scope>>();
+
 export const getScopeFromQuery = (
   schema: GraphQLSchema,
   query: string,
-): {
-  scope: NonNullable<CacheControlDirective['scope']>;
-  metadata: { privateProperty?: string };
-} => {
-  const fn = memoize1(({ query }: { query: string }) => {
+  options?: { sizePerSchema?: number },
+): Scope => {
+  if (!scopeCachePerSchema.has(schema)) {
+    scopeCachePerSchema.set(
+      schema,
+      new LRUCache({
+        max: options?.sizePerSchema ?? 1000,
+      }),
+    );
+  }
+
+  const cache = scopeCachePerSchema.get(schema);
+  const cachedScope = cache?.get(query);
+
+  if (cachedScope) return { ...cachedScope, metadata: { ...cachedScope.metadata, hitCache: true } };
+
+  function getScope() {
     const schemaCoordinates = getSchemaCoordinatesFromQuery(schema, query);
 
     for (const coordinate of schemaCoordinates) {
@@ -109,7 +128,10 @@ export const getScopeFromQuery = (
     }
 
     return { scope: 'PUBLIC' as const, metadata: {} };
-  });
+  }
 
-  return fn({ query });
+  const scope = getScope();
+  cache?.set(query, scope);
+
+  return scope;
 };
--- a/packages/plugins/response-cache/src/get-scope.ts
+++ b/packages/plugins/response-cache/src/get-scope.ts
@@ -1,0 +1,115 @@
+import {
+  FieldNode,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLOutputType,
+  GraphQLSchema,
+  Kind,
+  parse,
+  SelectionNode,
+  visit,
+} from 'graphql';
+import { memoize1 } from '@graphql-tools/utils';
+import { isPrivate, type CacheControlDirective } from './plugin';
+
+/** Parse the selected query fields */
+function parseSelections(selections: readonly SelectionNode[] = [], record: Record<string, any>) {
+  for (const selection of selections) {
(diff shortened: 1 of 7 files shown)


k-i-k-s:
Hi @ardatan, I'm back with a new proposal for this scope functionality with some improvements.
Tell me what you think about those changes!

ardatan:
@EmrysMyrddin Could you take a look?

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -558,13 +585,26 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
...
+              },
+              {
+                get: (obj, prop) => {
+                  if (prop === 'extras') {
+                    return getScopeFromQuery(schema, onExecuteParams.args.document.loc.source.body);
+                  }
+                  return obj[prop as keyof typeof obj];
+                },
+              },
+            ),
```
EmrysMyrddin: I don't remembre how it was implemented before, but is there a reason to use a proxy instead of classic getters?

Like this: 
```suggestion
            {
              documentString: getDocumentString(onExecuteParams.args),
              variableValues: onExecuteParams.args.variableValues,
              operationName: onExecuteParams.args.operationName,
              sessionId,
              context: onExecuteParams.args.contextValue,
              get extras() {
                return getScopeFromQuery(schema, onExecuteParams.args.document.loc.source.body);
              },
            },
```

packages/plugins/response-cache/src/get-scope.ts
```diff
@@ -0,0 +1,115 @@
...
+    for (const coordinate of schemaCoordinates) {
+      if (isPrivate(coordinate)) {
+        return { scope: 'PRIVATE' as const, metadata: { privateProperty: coordinate } };
+      }
+    }
+
+    return { scope: 'PUBLIC' as const, metadata: {} };
+  });
+
+  return fn({ query });
```
EmrysMyrddin: This will disallow memoization isn't it ? Since `{ query }` creates a new object each time the function is called.

EmrysMyrddin: And I'm not sure to understand why `schema` is not part of the `memoize1` key ? Because I fear that on schema update, you will get outadated metadata for the same query.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -295,20 +304,38 @@ const getDocumentWithMetadataAndTTL = memoize4(function addTypeNameToDocument(
...
 
-type CacheControlDirective = {
+export type CacheControlDirective = {
   maxAge?: number;
   scope?: 'PUBLIC' | 'PRIVATE';
 };
 
+export let schema: GraphQLSchema;
+let ttlPerSchemaCoordinate: Record<string, CacheControlDirective['maxAge']> = {};
+let scopePerSchemaCoordinate: Record<string, CacheControlDirective['scope']> = {};
```
EmrysMyrddin: Why do you extract those outside of the plugin? I feel that it will break things, because multiple instances of the same plugin would share those variables.

packages/plugins/response-cache/src/plugin.ts
```diff
@@ -295,20 +304,38 @@ const getDocumentWithMetadataAndTTL = memoize4(function addTypeNameToDocument(
...
+): boolean {
+  if (scopePerSchemaCoordinate[typeName] === 'PRIVATE') {
+    return true;
+  }
+  return data
+    ? Object.keys(data).some(
+        fieldName => scopePerSchemaCoordinate[`${typeName}.${fieldName}`] === 'PRIVATE',
+      )
+    : false;
+}
```
EmrysMyrddin: Why do you have to make a duplicate of this function outside of the plugin?

packages/plugins/response-cache/src/plugin.ts
k-i-k-s: A recent habit to use proxies for their flexibility, but I can see that a getter in this case would fit better for clarity sake.

packages/plugins/response-cache/src/get-scope.ts
k-i-k-s: Fair points!
I'll include `schema` using `memoize2` function and stringify those objects to fix comparison.

packages/plugins/response-cache/src/plugin.ts
k-i-k-s: Mb I merged local code I made last year with the one I use today and simply forgot there was already this function.
I'll keep the exported one since it's needed in `get-scope.ts` file.

k-i-k-s:
Hi @EmrysMyrddin, how's it going today ?

I made updates to fix the points you brought up :
- `extras` is a function now and the `schema` is its parameter
- no more exported `schema` since it's a param of `extras`
- memoization fixed with a combination of WeakMap (`schema` as the key) containing an LRU cache per schema (using `query` as key)
- also added `hitCache` boolean property in `metadata` and updated tests accordingly

k-i-k-s:
It seems square now with : 
- `options` parameter accessible to include metadata and alter size per schema in `LRUCache`
- changeset, unit tests and README updated

Tell me if there's still something blocking the merge @EmrysMyrddin.

Take care !

EmrysMyrddin:
Hi @k-i-k-s, would you mind allowing modification of your PR by maintainers?

I've taken a closer look and addressed issues so that we can finally merge this and handle the use case explained in the linked issue.

You can take a look at what it looks like here: https://github.com/graphql-hive/envelop/pull/2815

If you prefer not to give write access to your PR, I can merge mine instead.

k-i-k-s:
Hi, I don't think I have permission to give write access to the fork, merge your own branch I don't mind, it'll be far easier.
Good to see you're continuing implementation of this feature !
Is there an internal need on gql hive for you to personally intervene ?

EmrysMyrddin:
Ok, then don't hesitate to give my PR a try 🙂 

No, there isn't any new internal need for the feature, but I'm in the process of merging Envelop and Yoga repository together to ease development.

So before this can happen, we need to merge all PRs we don't want to loose. And I didn't want to let this one behind given the effort you put into it 🙂

PR #2217, opened 2024-04-26
[plugins/prometheus] Fix registry metrics duplication

EmrysMyrddin:
## Description

Is this plugin is initialized multiple times with the same registry (which can happen in the context of Mesh when polling is enabled for example).

This PR aims to fix this.

## Type of change

      expected)

## Further comments

Current solution rely on caching Histograms with a Map to reuse them instead of creating them each time.

But this has a drawback: if the configuration of those Histograms changes, this changement will be silently ignored and the old configuration will be kept in use.

### DIFF
--- a/.changeset/spotty-files-jam.md
+++ b/.changeset/spotty-files-jam.md
@@ -2,4 +2,83 @@
 '@envelop/prometheus': major
 ---
 
-TODO
+Adds a cache for metrics definition (Summary, Histogram and Counter).
+
+Fixes an issue preventing this plugin to be initialized multiple times, leading to metrics
+duplication error (https://github.com/ardatan/graphql-mesh/issues/6545).
+
+## Behavior Breaking Change:
+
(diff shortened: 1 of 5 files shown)


dotansimha:
> this changement will be silently ignored and the old configuration will be kept in use.
By "silent", you mean without any warning? should we at least warn or print something to the log? how silent it is?

dotansimha (approved):

EmrysMyrddin:
@dotansimha We don't actually have any means to know if the configuration have changed or not. Even if we keep track of the previous configuration somehow and check fo the deep equality with the new config, it will most likely be always different, because of the `fillLabels` being a function, which will always be uniq even if it has the exact same implementation.

We could add a warning if we detect that a metrics was already registered with this name, but then will have this warning every time we create a new Mesh instance when polling is enable.

Do you think I should add a warning with an option to disable it explicitly ?

PR #2218, opened 2024-04-26
fix lockfile for newrelic tests

EmrysMyrddin:
Revert the lockfile to it's state before #eafae3bf93a19f2e41efcadcc1c42fb99ba08b52.

This commit has entirely rewritten the lockfile, leading to unmanaged dependencies updates breaking NewRelic plugin tests.

### DIFF
--- a/packages/plugins/opentelemetry/package.json
+++ b/packages/plugins/opentelemetry/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@envelop/on-resolve": "^4.1.0",
-    "@opentelemetry/api": "^1.0.0",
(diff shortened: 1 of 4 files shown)


packages/plugins/opentelemetry/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@envelop/on-resolve": "^4.1.0",
-    "@opentelemetry/api": "^1.0.0",
```
EmrysMyrddin, line 55: It seems that we are using some newer stuff than `1.0.0`.
Not sure what the minimal version should be, so updating to the latest.

PR #2230, opened 2024-05-08
use `workspace:*` and update pnpm

dotansimha:


EmrysMyrddin:
Actually, perhaps we will need this... Because it's now broken with pnpm 9

Issue #2104, opened 2023-12-05
[CloudflareKVPlugin]: Incompatible types when passing in argument for kv


AdiRishi:
**Describe the bug**

```ts
      const kvCache = createKvCache({
        KV: env.GRAPHQL_RESPONSE_CACHE,
        ctx,
        keyPrefix: 'graphql', // optional
        cacheReadTTL: 1000 * 60 * 60 * 4, // 4 hours
      });
```

When using the `createKvCache` export from `@envelop/response-cache-cloudflare-kv` the passed in type of `KV` is rejected with this error

```
src/index.ts:29:9 - error TS2322: Type 'KVNamespace<string>' is not assignable to type 'import("/Users/arishi/personal/sukrit-ecosystem-turborepo/node_modules/.pnpm/@cloudflare+workers-types@4.20231121.0/node_modules/@cloudflare/workers-types/index").KVNamespace<string>'.
  Types of property 'get' are incompatible.
    Type '{ (key: string, options?: Partial<KVNamespaceGetOptions<undefined>> | undefined): Promise<string | null>; (key: string, type: "text"): Promise<...>; <ExpectedValue = unknown>(key: string, type: "json"): Promise<...>; (key: string, type: "arrayBuffer"): Promise<...>; (key: string, type: "stream"): Promise<...>; (key:...' is not assignable to type '{ (key: string, options?: Partial<import("/Users/arishi/personal/sukrit-ecosystem-turborepo/node_modules/.pnpm/@cloudflare+workers-types@4.20231121.0/node_modules/@cloudflare/workers-types/index").KVNamespaceGetOptions<undefined>> | undefined): Promise<...>; (key: string, type: "text"): Promise<...>; <ExpectedValue ...'.
      Types of parameters 'options' and 'type' are incompatible.
        Type '"stream"' has no properties in common with type 'Partial<KVNamespaceGetOptions<undefined>>'.

29         KV: env.GRAPHQL_RESPONSE_CACHE,
           ~~

  ../../node_modules/.pnpm/@envelop+response-cache-cloudflare-kv@0.2.0_@envelop+response-cache@6.1.1_graphql@16.8.1/node_modules/@envelop/response-cache-cloudflare-kv/typings/index.d.ts:7:5
    7     KV: KVNamespace;
          ~~
    The expected type comes from property 'KV' which is declared here on type 'KvCacheConfig'


Found 1 error in src/index.ts:29
```

**To Reproduce** Steps to reproduce the behavior:

[...]

PR #2237, opened 2024-05-18
Pin to the latest compatibility date for Cloudflare workers

AdiRishi:
This change applies to the kv-response-cache plugin

## Description

The `@cloudflare/worker-types` package has multiple versions of its types that are pinned to different compatibility dates. The When you create a new cloudflare project using their CLI, they now default to the latest date whcih is `2023-07-01`.

This PR updates the imports for the kv response cache plugin to use the name compatibility dates as the default for new projects, this should incraese compatibility for users.

For issue #2104 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Testing this is a bit challenging. The types are correct locally, the hope is that when this package is imported there will be fewer conflicts with types.

## Checklist:

- [X] I have followed the
- [X] I have performed a self-review of my own code
- [X] I have commented on my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -1,5 +1,5 @@
 import type { ExecutionResult } from 'graphql';
-import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types';
+import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types/2023-07-01';
 import type { Cache, CacheEntityRecord } from '@envelop/response-cache';
 import { buildOperationKey } from './cache-key.js';
 import { invalidate } from './invalidate.js';
--- a/.changeset/healthy-hotels-design.md
+++ b/.changeset/healthy-hotels-design.md
@@ -1,0 +1,5 @@
+---
+'@envelop/response-cache-cloudflare-kv': minor
+---
+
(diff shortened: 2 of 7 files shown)


EmrysMyrddin:
Great ! Thank you for the PR!
Does this fix all the types issues we had ?

packages/plugins/response-cache-cloudflare-kv/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import type { ExecutionResult } from 'graphql';
-import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types';
+import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types/2023-07-01';
```
EmrysMyrddin, line 2: Doesn't this have the potential to age poorly ? We will have to update all imports when we update the types dependency ?

EmrysMyrddin:
@AdiRishi We also have another PR (#2238) which should even remove the necessity to have a type matching on user land. This PR doesn't take any CFW types anymore and takes the namespace name as string instead. Which both simplify the usage and should fix the type issue :-) 

But I still think we should pin the typing anyway

PR #2238, opened 2024-05-20
enhance(response-cache/cfw-kv): simplify cache init

ardatan:
This PR introduces the following changes;

- Now `cache` parameter accepts a factory function that takes the context and returns the cache implementation
- Now the cache implementation does not require any ctx value, but only the name of the KV namespace.

See the changes in the README for details

```ts
import { createSchema, createYoga, YogaInitialContext } from 'graphql-yoga'
import { useResponseCache } from '@envelop/response-cache'
import { createKvCache } from '@envelop/response-cache-cloudflare-kv'
import { resolvers } from './graphql-schema/resolvers.generated'
import { typeDefs } from './graphql-schema/typeDefs.generated'

export type Env = {
  GRAPHQL_RESPONSE_CACHE: KVNamespace
}

const graphqlServer = createYoga<Env & ExecutionContext>({
  schema: createSchema({ typeDefs, resolvers }),
  plugins: [
    useResponseCache({
      cache: createKvCache({
        KVName: 'GRAPHQL_RESPONSE_CACHE',
        keyPrefix: 'graphql' // optional
      }),
      session: () => null,
      includeExtensionMetadata: true,
      ttl: 1000 * 10 // 10 seconds
    })
  ]
})

export default {
  fetch: graphqlServer
}
```

### DIFF
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -110,7 +110,7 @@ export function createKvCache<
           return invalidatePromise;
         }
         // Do not block execution of the worker while invalidating the cache
-        ctx.waitUntil(invalidate(entities, ctx[config.KVName], config.keyPrefix));
+        ctx.waitUntil(invalidatePromise);
       },
     };
   };
--- a/packages/plugins/response-cache-cloudflare-kv/src/index.ts
+++ b/packages/plugins/response-cache-cloudflare-kv/src/index.ts
@@ -1,5 +1,5 @@
 import type { ExecutionResult } from 'graphql';
-import type { ExecutionContext, KVNamespace } from '@cloudflare/workers-types';
+import type { KVNamespace } from '@cloudflare/workers-types';
 import type { Cache, CacheEntityRecord } from '@envelop/response-cache';
 import { buildOperationKey } from './cache-key.js';
 import { invalidate } from './invalidate.js';
@@ -11,9 +11,9 @@ export type KvCacheConfig = {
    */
   KV: KVNamespace;
   /**
-   * The Cloudflare worker execution context. Used to perform non-blocking actions like cache storage and invalidation.
+   * The function that should be used to wait for non-blocking actions to complete
    */
-  ctx: ExecutionContext;
+  waitUntil: (promise: Promise<unknown>) => void;
   /**
    *  Defines the length of time in milliseconds that a KV result is cached in the global network location it is accessed from.
    *
@@ -67,12 +67,12 @@ export function createKvCache(config: KvCacheConfig): Cache {
       ttl: number,
     ) {
       // Do not block execution of the worker while caching the result
-      config.ctx.waitUntil(set(id, data, entities, ttl, config));
+      config.waitUntil(set(id, data, entities, ttl, config));
     },
 
     invalidate(entities: Iterable<CacheEntityRecord>) {
       // Do not block execution of the worker while invalidating the cache
-      config.ctx.waitUntil(invalidate(entities, config));
+      config.waitUntil(invalidate(entities, config));
     },
   };
   return cache;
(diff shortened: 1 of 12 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -550,7 +553,8 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           return;
         }
 
-        cache.set(cacheKey, result, identifier.values(), finalTtl);
+        const cacheInstance = cacheFactory(onExecuteParams.args.contextValue);
+        cacheInstance.set(cacheKey, result, identifier.values(), finalTtl);
```
EmrysMyrddin: Why do you instanciate again here ? You should still have access to the `cacheInstance` declared in the upper scope on l.529 ?

EmrysMyrddin:
Just to let you know, there is another PR (https://github.com/n1ru4l/envelop/pull/2237) which also updates the CFW cache plugin. It updates types, so it should not be of a huge impact, I probably can do the rebase for you after merging if you want.

ardatan:
You can rebase but also keep in mind, my PR removes the need of the type compatibility. So we don't even need to deal with ReadacleStream conflict in that case.

EmrysMyrddin:
We are still importing the `KVNamespace` type aren't we ?

You mean that the fact we don't take this as a parameter anymore avoid the typing issue in user land ?

ardatan:
@EmrysMyrddin I believe so but even if not, that is not the main goal of this PR. The main goal is to simplify the usage, move the server definition etc on top, less verbose API etc.

packages/plugins/response-cache/src/plugin.ts
ardatan: Removed!

EmrysMyrddin (approved):

EmrysMyrddin:
I still think it can be good to merge the other PR anyway, pinning version could avoid breaking the plugin on updates ?

EmrysMyrddin:
@ardatan Is it ok to merge this ?

ardatan:
Sure you can rebase the conflicts and merge!

packages/plugins/response-cache-cloudflare-kv/src/index.ts
```diff
@@ -44,36 +47,71 @@ export function createKvCache(config: KvCacheConfig): Cache {
...
+        const invalidatePromise = invalidate(entities, ctx[config.KVName], config.keyPrefix);
+        if (!ctx.waitUntil) {
+          // eslint-disable-next-line no-console
+          console.warn(
+            'The server context does not have a waitUntil method. This means that the cache invalidation will not be non-blocking.',
+          );
+          return invalidatePromise;
+        }
+        // Do not block execution of the worker while invalidating the cache
+        ctx.waitUntil(invalidate(entities, ctx[config.KVName], config.keyPrefix));
```
EmrysMyrddin: You are invalidating 2 times I think
```suggestion
        ctx.waitUntil(invalidatePromise);
```

packages/plugins/response-cache-cloudflare-kv/src/index.ts
```diff
@@ -44,36 +47,71 @@ export function createKvCache(config: KvCacheConfig): Cache {
...
-      const kvResponse = await config.KV.get(buildOperationKey(id, config.keyPrefix), {
-        type: 'text',
-        cacheTtl: computedTtlInSeconds,
-      });
-      if (kvResponse) {
-        return JSON.parse(kvResponse) as ExecutionResult;
-      }
-      return undefined;
-    },
+  return function KVCacheFactory(ctx: TServerContext): Cache {
```
EmrysMyrddin, line 50: We can probably even warn only once at the start of the function isn't it ?

```suggestion
  return function KVCacheFactory(ctx: TServerContext): Cache {
    if (!ctx[config.KVName]) {
      // eslint-disable-next-line no-console
      console.warn(
        `Cloudflare KV namespace ${config.KVName} is not available in the server context, cache will be skipped.`,
      );
      return {
        get: () => undefined,
        set: () => undefined,
        invalidate: () => undefined,
      };
    }
```

ardatan: I don't know the lifecycle of CF Workers, so I thought it might be added later on after startup. I preferred to check it lazily.

PR #2241, opened 2024-05-22
[useResponseCache] Ensure calls to cache are awaited

### LINKED ISSUE #2240: [useResponseCache] Promise rejections from cache impl are not handled
### Issue workflow progress

- [ ] 1. The issue provides a
     [Stackblitz](https://stackblitz.com/fork/node).
- [x] 2. A failing test has been provided
- [x] 3. A local solution has been provided
- [x] 4. A pull request is pending review

---

**Describe the bug**

The cache implementation can be sync or async - each of the methods can return a Promise. However, these promises are not awaited or handled so they end up as "floating promises". When one of these rejects, Node treats it as an unhandled promise rejection and, under default behaviour

1. Set up [response cache with graphql-yoga](https://the-guild.dev/graphql/yoga-server/docs/features/response-caching) in Node.
2. Override the cache with an implementation of `set` which returns a rejected promise.
3. Execute a query that would be cached.
4. The result returns since `cache.set` isn't awaited, but in the background the Node processes crashes.

We noticed this whilst using `@envelop/response-cache-redis` with a Redis setup that wasn't configured correctly, however, the incorrect error handling can be reproduced with anything that returns a rejected promise.

**Expected behavior**

The rejected promise should be returned through `onExecuteDone` so it can be handled by the server and result in a failed request. The process should not get terminated. 

**Environment:**

- OS: macOS 14.4.1
- NodeJS: v20.11.0
- `@envelop/*` versions:
  - `@envelop/response-cache`: `6.1.2`

sachinahya:
## Description

This change ensures that the calls to the supplied cache implementation are awaited and returned, in the case that the methods are async and return promises. This prevents "floating promises" from being created which will crash the Node process if they are rejected.

I believe this is not a breaking change since all supported Node versions have the behaviour of terminating on unhandled promise rejections, and that the behaviour without this change is not intentional or desired.

Fixes #2240.

## Type of change

## How Has This Been Tested?

Unit tests added for `set` and `invalidate` calls to the cache implementation, ensuring that sync errors and promise rejections don't result in process termination.

**Test Environment**:

- OS: macOS 14.4.1
- `@envelop/response-cache`: 6.1.2
- NodeJS: v20.11.0

## Checklist:

- [x] I have followed the
- [ ] My changes generate no new warnings

### DIFF
--- a/.changeset/old-moles-develop.md
+++ b/.changeset/old-moles-develop.md
@@ -1,0 +1,6 @@
+---
+'@envelop/response-cache': patch
+---
+
+Fixes an issue where rejected promises returned from the cache were not being handled correctly,
+causing an unhandled promise rejection which would terminate the process.
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -472,13 +472,13 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
         }
       }
 
-      function invalidateCache(
+      async function invalidateCache(
         result: ExecutionResult,
         setResult: (newResult: ExecutionResult) => void,
-      ): void {
+      ): Promise<void> {
         processResult(result.data);
 
-        cache.invalidate(identifier.values());
+        await cache.invalidate(identifier.values());
         if (includeExtensionMetadata) {
           setResult(
             resultWithMetadata(result, {
@@ -535,7 +535,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
         });
       }
 
-      function maybeCacheResult(
+      async function maybeCacheResult(
         result: ExecutionResult,
         setResult: (newResult: ExecutionResult) => void,
       ) {
@@ -550,7 +550,7 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
           return;
         }
 
-        cache.set(cacheKey, result, identifier.values(), finalTtl);
+        await cache.set(cacheKey, result, identifier.values(), finalTtl);
         if (includeExtensionMetadata) {
           setResult(resultWithMetadata(result, { hit: false, didCache: true, ttl: finalTtl }));
         }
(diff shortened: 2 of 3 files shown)


EmrysMyrddin:
Thank you for you contribution! Very appreciated!

I agree that we should not have floating promises without any error handling here, but I'm not sure that awaiting it is the right approach about this.

My concern is that if there is a `set` failure, the entire request handling will fail. You probably don't want to fail responding a valid request just because the cache is down, and you probably don't want to add the latency of the cache network travel.

For this reasons, perhaps just adding a `.catch(err => logger.warn('Failed to cache response', err))` instead ?
Do you think it would suite your needs ?

The drawback is that it opens the possibility to have hanging open handles when shuting down the server, which can prevent server for gracefully exit. That being said, it's already the case with the current implementation. We should probably better add an abort signal to allow canceling these promises on shutdown, but not sure it is worth the added complexity for a case that is probably very rare.

EmrysMyrddin:
After some discussion with a colleague, it appears that we actually already have a way to handle floating promises like this.
There is an undocumented `waitUntil` function in the yoga context.
It's provided by `whatwg-node/server`, and it's signature is `waitUntil(f: Promise<void> | void): void`.

sachinahya:
Adding a `catch` handler works for us too. I also agree that the case for aborting floating promises is not worth the complexity since I'd imagine it's very rare to have a promise that never resolves.

As far as logging goes, I'm not sure where `logger` would come from in your example as I can't see where it gets passed in from. Maybe we can add an empty `.catch` and leave the logging to the implementation of `Cache`, or add a `console.error`?

EmrysMyrddin:
Sorry, I was not very clear.
In fact, there is a function in the context called `waitUntil` that is already doing exactly this :-) And it allows for a better compatibility in ephemeral environment like cloud functions. You can use it instead of manually adding a `.catch`.
Under the hood, if an implementation is not provided by the environment, it calls `.allSettled` on every promises passed to it as argument.

sachinahya:
Thanks, could you point us to some documentation for this once it's ready and we'll give it a go?

PR #2249, opened 2024-06-06
[useResponseCache] Ignore specfic schemaCoordinates for IdFields which are not the root of a subgraph

### LINKED ISSUE #2247: [useResponseCache] Ignore specfic schemaCoordinates for IdFields which are not the root of a subgraph

In our schema we have fields named "id" which are not part of a subgraph, just a nested field named "id".  This causes the caching-plugin to treat this nested field like a subgraph, which results in the nested field always beeing null. 

Add a new  optionalList of schemaCoordinates to the useResponseCacheParamters, which is used to exclude fields on specific objects in the mapSchema step. Can provide a pull request. 

Alternative solution which we tried (and worked) was renaming all "id" fields which are not a root of a subgraph in our schema. Which sadly is not a solution wie can deploy.

Kendalor:
🚨 **IMPORTANT: Please do not create a Pull Request without creating an issue first.**

the pull request._

## Description

Add a new optionalList of schemaCoordinates to the useResponseCacheParamters, which is used to exclude fields on specific objects in the mapSchema step.

Fixes #2247

## Type of change

N/A

## How Has This Been Tested?

Testet fix first on our case and environment by maintaining a parallel instance of this plugin. Further added Unit tests to first reproduce the Issue and proof the fix works. 

- [x] Added Test "'id field in body overwritten and request fails'"
- [x] Added Test "'id field in body is returned as is and not overwritten'"

**Test Environment**:

- OS: macOs: Sonoma 14.5
- `@envelop/response-cache`: 6.2.0
- NodeJS: 10.5.0

## Checklist:

- [ x] I have followed the
- [ x] I have performed a self-review of my own code

## Further comments

N/A

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -115,6 +115,11 @@ export type UseResponseCacheParameter<PluginContext extends Record<string, any> 
    */
   idFields?: Array<string>;
   /**
+   * List of SchemaCoordinates which are ignored during scan for entities.
+   * Defaults to `[]`
+   */
+  ignoreIdFieldsBySchemaCoordinate?: Array<string>
+  /**
    * Whether the mutation execution result should be used for invalidating resources.
    * Defaults to `true`
    */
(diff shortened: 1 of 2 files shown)


Kendalor:
@ardatan  is anythign missing ? Should the changeset be created by me ?

ardatan:
@EmrysMyrddin Could you review?

EmrysMyrddin:
Hi, I'm not entirely against this feature, but I'm afraid we are introducing it to workaround an actual bug.
I'm not sure to entirely understand what is happening here. We should probably first investigate why the plugin behave like this.

Kendalor:
@EmrysMyrddin  how would you suggest to move forward ?
From my expierience/debugging suggestion the plugin identifies every object with an "id" field (configurable) as a cachable entity from a subgraph.  If Subgraph A returns a value containing a field with an "id", the plugin seems to remove this part of the resonse while stitching the response together from a non existing subgraph B. Therefore setting the field to null.

Issue #2256, opened 2024-06-22
Apollo client support (community contribution)


vladinator1000:
Hi there, I have a [server-side rendering use-case](https://github.com/n1ru4l/envelop/discussions/2250) and I wrote a thing for it. It lets you use Envelop with Apollo client.

## Example
```ts
import * as graphqlJs from 'graphql'
import { ApolloClient, gql, InMemoryCache } from '@apollo/client'
import { EnvelopSchemaLink } from '@envelop/apollo-client'
import { envelop, useEngine, useSchema } from '@envelop/core'
import { makeExecutableSchema } from '@graphql-tools/schema'

let schema = makeExecutableSchema({
  typeDefs: `type Query { hello: String! }`,
  resolvers: {
    Query: {
      hello: () => 'world'
    }
  }
})

// Use any enveloped setup
let getEnveloped = envelop({
  plugins: [useEngine(graphqlJs), useSchema(schema)]
})
let envelope = getEnveloped()

let apollo = new ApolloClient({
  cache: new InMemoryCache(),
  // Pass it to EnvelopSchemaLink
  link: new EnvelopSchemaLink(envelope)
})

// Use Apollo
let result = await apollo.query({
  query: gql`
    query {
      hello
    }
  `
})

console.log(result)
```

## Contribution

Here's the diff, would you accept a PR similar to this?
https://github.com/n1ru4l/envelop/compare/main...vladinator1000:envelop:apollo-schema-link?expand=1

EmrysMyrddin:
Wow that looks awesome !
A PR would be very welcomed !

vladinator1000:
@EmrysMyrddin 🚀 https://github.com/n1ru4l/envelop/pull/2257

PR #2257, opened 2024-06-24
Apollo Client - schema link integration

vladinator1000:
## Description

Useful for server-side rendering that avoids network calls, if you're rendering on the same server that the GraphQL API runs on.

https://github.com/n1ru4l/envelop/issues/2256

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)

## How Has This Been Tested?

- [x] Unit test (packages/plugins/apollo-client/test/envelopSchemaLink.spec.ts)

## Checklist:

- [x] I have followed the


packages/plugins/apollo-client/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ComposeContext, GetEnvelopedFn, Optional, Plugin } from '@envelop/types';
+
+type ExcludeFalsy<TArray extends any[]> = Exclude<TArray[0], null | undefined | false>[];
+
+export namespace EnvelopSchemaLink {
+  export type ResolverContext = Record<string, any>;
+  export type ResolverContextFunction = (
+    operation: Operation,
+  ) => ResolverContext | PromiseLike<ResolverContext>;
+  export type Options<PluginsType extends Optional<Plugin<any>>[]> = ReturnType<
```
vladinator1000, line 11: Types copied from https://github.com/n1ru4l/envelop/blob/main/packages/core/src/create.ts#L10.

Plugin inspired by https://github.com/apollographql/apollo-client/blob/main/src/link/schema/index.ts#L8

packages/plugins/apollo-client/test/envelopSchemaLink.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      query {
+        hello
+      }
+    `;
+
+    await apollo.query({
+      query,
+    });
+
+    let mockCallResult = logMock.mock.calls[1][1].result;
```
vladinator1000, line 49: Asserting the second logger call ("query-end"), because it contains the resolved data.

packages/plugins/apollo-client/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+          if (validationErrors.length > 0) {
+            return { errors: validationErrors };
+          }
+
+          return this.envelope.execute({
+            schema: this.envelope.schema,
+            document: operation.query,
+            rootValue: {
+              execute: this.envelope.execute,
+              subscribe: this.envelope.subscribe,
```
vladinator1000, line 45: I'm not sure what to put here. [Lee Byron says](https://github.com/graphql/graphql-js/issues/735#issuecomment-302456482) 

>The root value represents the "top" of your metaphorical graph of data, and is useful to include functions or data to help resolve the root fields in your schema.

I've personally never had the need to use `rootValue`. It seems like Apollo's SchemaLink [has it as an option](https://github.com/apollographql/apollo-client/blob/main/src/link/schema/index.ts#L47)

examples/apollo-client/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      hello: () => 'world',
+    },
+  },
+});
+
+// Use any enveloped setup
+const getEnveloped = envelop({
+  plugins: [useEngine(graphqlJs), useSchema(schema)],
+});
+const envelope = getEnveloped();
```
ardatan, line 20: I think `getEnveloped` should be called per operation instead of once.
Because the lifecycle in Envelop is handled per operation. So it would be better to call it inside the link.

examples/apollo-client/index.ts
```diff
@@ -0,0 +1,42 @@
+import * as graphqlJs from 'graphql';
+import { ApolloClient, gql, InMemoryCache } from '@apollo/client';
+import { EnvelopSchemaLink } from '@envelop/apollo-client';
```
ardatan, line 3: Let's call it only `EnvelopLink`.
EnvelopSchema seems extra :)

packages/plugins/apollo-client/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+          const validationErrors = this.envelope.validate(this.envelope.schema, operation.query);
+
+          if (validationErrors.length > 0) {
+            return { errors: validationErrors };
+          }
+
+          return this.envelope.execute({
+            schema: this.envelope.schema,
+            document: operation.query,
+            rootValue: {
```
ardatan, line 43: We don't need this I guess.

packages/plugins/apollo-client/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+        resolve(this.envelope.contextFactory),
+      )
+        .then(context => {
+          const validationErrors = this.envelope.validate(this.envelope.schema, operation.query);
+
+          if (validationErrors.length > 0) {
+            return { errors: validationErrors };
+          }
+
+          return this.envelope.execute({
```
ardatan, line 40: Do you think we can also support `subscribe` ?
Also keep on mind that `execute` can return AsyncIterable.

ardatan: This can be similar to this one;
https://github.com/ardatan/graphql-mesh/blob/master/packages/legacy/apollo-link/src/index.ts

PR #2258, opened 2024-06-28
[response-cache] deprecate `ttlPerType` option in favor of `ttlPerSchemaCoordinate`

EmrysMyrddin:
## Description

There is an incoherence in the options API. TTL is configured using 2 parameters: `ttlPerType` and `ttlPerSchemaCoordinate` while scope is configured only via `scopeBySchemaCoordinate` which can contain typeName only or typeName+fieldName keys.

This PR is deprecating the `ttlPerType` in favor of `ttlPerSchemaCoordinate` to be more coherent.
`ttlPerType` is still supported but will be removed in future major release.

## Type of change

### DIFF
--- a/.changeset/twenty-camels-wonder.md
+++ b/.changeset/twenty-camels-wonder.md
@@ -1,0 +1,25 @@
+---
+'@envelop/response-cache': minor
+---
+
+Deprecate `ttlPerType` in favor of `ttlPerSchemaCoordinate`, for a more streamlined API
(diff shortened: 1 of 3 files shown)


dotansimha (approved):

n1ru4l (approved):

EmrysMyrddin:
I just realized that I didn't updated tests. I have to do this before merging.

PR #2270, opened 2024-07-26
refactor(prometheus-plugin): rename metrics options

EmrysMyrddin:
## Description

For this plugin to work, each metric the user want to be exported should be manually enabled using the corresponding option. But the name of the exported metric and its corresponding option are different, which is confusing for users.

This PR renames all the options to use their actual metric name.

## Type of change

      expected)

### DIFF
--- a/.changeset/heavy-houses-kick.md
+++ b/.changeset/heavy-houses-kick.md
@@ -1,0 +1,51 @@
+---
+'@envelop/prometheus': major
+---
+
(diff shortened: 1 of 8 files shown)


EmrysMyrddin:
The last commit is a suggestion, perhaps it could be better to separate metrics enabling/disabling options from the rest, just as we do for labels.

It looks better for types and will be easier to explain in documentation I think.

dotansimha (approved):

dotansimha (approved):

PR #2276, opened 2024-08-01
Add support for all graphql-rate-limit options

### LINKED ISSUE #2273: Add store support to rate-limiter

We're trying to implement rate-limiting on our graphql-yoga server, but we need the store to use Redis. Currently, it appears that we are simply missing a pass through from the `graphql-rate-limit` package and this should be a simple addition.

Support the `store` prop on `useRateLimiter` so that Redis can be used as a store option.

Go back to using `graphql-rate-limit` and moving away from `graphql-yoga`. Unfortunately, we would lose other features that we need that are included with `graphql-yoga`.

**Additional context**

I've started making the code changes necessary for a PR here:
https://github.com/deggertsen/envelop/blob/feat/add-rate-limiter-store-support/packages/plugins/rate-limiter/src/index.ts
I'm not sure if we also need/want to export the Store and RedisStore types in addition to adding the store prop to `useRateLimiter`.

deggertsen:
## Description

Exposes all possible configuration fields and types to the rate-limiter plugin and forwards them to the underlying graphql-rate-limit function. This is as discussed in issue #2273. I've designed it to be backwards compatible so that it shouldn't break past versions, but all config should now be available through the graphQLRateLimitConfig prop.

Primary motivation is to expose the "store" prop so that Redis can be used.

Fixes #2273

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as
      expected)

## How Has This Been Tested?

This is a fairly simple change that shouldn't be breaking. I've tested locally by copying the plugin to my project and using the Redis store option. It seems to be working as expected.

**Test Environment**:

- OS: macOS 14.6
- `@envelop/rate-limiter`: 6.1.0
- NodeJS: 18.17.0

## Checklist:

- [x] I have followed the

### DIFF
--- a/packages/plugins/rate-limiter/src/index.ts
+++ b/packages/plugins/rate-limiter/src/index.ts
@@ -45,7 +45,7 @@ interface RateLimiterContext {
 
 export const useRateLimiter = (options: RateLimiterPluginOptions): Plugin<RateLimiterContext> => {
   const rateLimiterFn = getGraphQLRateLimiter({
-    ...options.graphQLRateLimitConfig,
+    ...options.graphQLRateLimitConfig, // Pass through all available options
     identifyContext: options.identifyFn,
   });
 
--- a/packages/plugins/rate-limiter/src/index.ts
+++ b/packages/plugins/rate-limiter/src/index.ts
@@ -6,6 +6,18 @@ import { getDirective } from './utils.js';
 
 export * from './utils.js';
 
+export {
+  FormatErrorInput,
+  GraphQLRateLimitConfig,
+  GraphQLRateLimitDirectiveArgs,
+  Identity,
+  InMemoryStore,
+  Options,
+  RateLimitError,
+  RedisStore,
+  Store,
+} from 'graphql-rate-limit';
+
 export class UnauthenticatedError extends Error {}
 
 export type IdentifyFn<ContextType = unknown> = (context: ContextType) => string;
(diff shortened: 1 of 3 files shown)


packages/plugins/rate-limiter/src/index.ts
```diff
@@ -24,14 +36,18 @@ export type RateLimiterPluginOptions = {
     context: unknown;
     info: GraphQLResolveInfo;
   }) => void;
+  graphQLRateLimitConfig?: GraphQLRateLimitConfig;
 };
```
EmrysMyrddin: Perhaps we can just join the 2 types ? This way, all options are directly available, it's better for API discovery for users that don't have pre-existing experience of `graphql-rate-limit` :-)

We just have to make sure to omit the option that we have renamed (not sure why we did this in the first place, but not a big deal, not sure it is worth the breaking change to rename it)

```suggestion
} & Omit<GraphQLRateLimitConfig, 'identifyContext'>;
```

EmrysMyrddin (commented):
Thank you for your contribution! Very appreciated :-)

I've made a change proposition, feel free to discuss it.

Can you also please add a changeset? It's our tool to manage changelogs and versioning of the packages across the mono-repo. You can just run `pnpm changeset` and follow the instructions :-)

deggertsen:
I merged in your suggestion and fixed the spread of options to reflect that change.

I also added a changeset. Let me know if this should be a minor release instead of a patch release, but I think a patch seems right since there aren't any breaking changes.

Thanks for the feedback and recommendations! Let me know if there is anything else.

PR #2309, opened 2024-10-09
fix: strip response cache fields from incremental execution result

### LINKED ISSUE #2308: response-cache does not clear cache info from response for incremental delivery
### Issue workflow progress

When using the response cache plugin with incremental delivery (return stream from execute), the `__responseCacheId` and `__responseCacheTypeName` fields are not cleared from the parts.

image

n1ru4l:
closes https://github.com/n1ru4l/envelop/issues/2308

### DIFF
--- a/packages/plugins/response-cache/src/plugin.ts
+++ b/packages/plugins/response-cache/src/plugin.ts
@@ -698,12 +698,8 @@ function removeMetadataFieldsFromResult(data: Record<string, unknown>): any {
   // clone the data to avoid mutation
   data = { ...data };
 
-  if (data.__responseCacheTypeName) {
-    delete data.__responseCacheTypeName;
-  }
-  if (data.__responseCacheId) {
-    delete data.__responseCacheId;
-  }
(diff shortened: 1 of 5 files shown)


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -684,3 +690,26 @@ export const cacheControlDirective = /* GraphQL */ `
...
+  }
+  // clone the data to avoid mutation
+  data = { ...data };
+
+  if (data.__responseCacheTypeName) {
+    delete data.__responseCacheTypeName;
+  }
+  if (data.__responseCacheId) {
+    delete data.__responseCacheId;
+  }
```
EmrysMyrddin: It is safe to delete things even if they don't exists
```suggestion
  delete data.__responseCacheTypeName;
  delete data.__responseCacheId;
```

EmrysMyrddin (approved):

dotansimha (approved):

PR #2310, opened 2024-10-09
fix: add type annotations for function and make it typesafe

n1ru4l:


EmrysMyrddin (approved):
I'm not against this change, but just for the context, the code was like this because we wanted to not adhere to any graphql engine. Which arguably, we already are...

dotansimha (approved):

PR #2317, opened 2024-10-11
feat: [plugin/prometheus] allow to skip observation based on context

### LINKED ISSUE #2312: `[@envelop/plugin-prometheus]` Need a way to filter which metrics are recorded

In our federated setup we have to be restrictive on histogram metrics, our we will have too much metrics too scrape. To some extent this can be fixed by customizing labels and reducing cardinality, but in some cases we actually want to opt in to the high cardinality metrics for a selection of operations.

Thus, we would like to have e.g. `graphql_yoga_http_duration` enabled but only if `operationName` matches certain allow listed values.

We would like to be able to have a (mesh) config like this in the end

```ts
{
  prometheus: {
    metrics: {
      graphql_yoga_http_duration: createHistogram({
        registry,
        histogram: {
          name: 'graphql_yoga_http_duration',
          help: 'Time spent on HTTP connection',
          labelNames: ['operation_name'],
          buckets: HISTOGRAM_BUCKET_DIMENSIONS,
        },
        fillLabelsFn: ({ operationName }, _rawContext) => ({
          operation_name: operationName,
        }),
        shouldRecordFn: ({ operationName }, _rawContext) =>
          TRACKED_OPERATIONS.includes(operationName),
      })
    }
  }
}
```

Have to build our own stack of plugins all the way down to envelop (mesh plugin, yoga plugin, envelop plugin).

EmrysMyrddin:
## Description

This PR aims to give more control over exposed metrics.

Today, the configuration is very confusing about what metrics are enabled or not, which can lead to silently broken configuration.

I propose a new way to configure which metrics should be enabled, for which execution phases, with which labels. The goal is also to offer a more dynamic approach by giving a way to skip metric observation based on execution context.

Fixes #2312

## Type of change

### DIFF
--- a/packages/plugins/prometheus/README.md
+++ b/packages/plugins/prometheus/README.md
@@ -72,6 +72,11 @@ Histogram metrics can be passed an array of numbers to configure buckets.
 Each metric also expose a set of labels. All labels are exposed by default but can be separately
 disabled by setting the corresponding key in `labels` option object to `false`.
 
+A metric can observe events in different phases of of GraphQL request pipeline. By default, if a
+metric is available, it will observe timing or events in every available phases for this metric. You
+can configure this by either providing an array instead of `true` in the metrics config, or use the
+`phases` option in the custom metric factory.
+
 ### `graphql_envelop_phase_parse`
 
 This metric tracks the duration of the `parse` phase of the GraphQL execution. It reports the time
@@ -519,6 +524,7 @@ const getEnveloped = envelop({
             help: 'HELP ME',
             labelNames: ['opText'] as const,
           }),
+          phases: ['parse'], // This is an array of phases that should be hooked for this metric
           fillLabelsFn: params => {
             // if you wish to fill your `labels` with metadata, you can use the params in order to get access to things like DocumentNode, operationName, operationType, `error` (for error metrics) and `info` (for resolvers metrics)
             return {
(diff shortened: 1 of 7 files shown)


dotansimha (approved):

EmrysMyrddin:
@klippx Hey! What do you think of this new configuration API ? Would it cover you needs ?
You can see the README changes to see how to use it :-)

packages/plugins/prometheus/README.md
```diff
@@ -532,6 +538,73 @@ const getEnveloped = envelop({
...
+          registry,
+          histogram: {
+            name: 'graphql_yoga_http_duration',
+            help: 'Time spent on HTTP connection',
+            labelNames: ['operation_name']
+          },
+          fillLabelsFn: ({ operationName }, _rawContext) => ({
+            operation_name: operationName
+          }),
+          shouldObserve: context => TRACKED_OPERATIONS.includes(context?.params?.operationName)
```
klippx, line 600: I like it! 

But I think the example should be updated:

```suggestion
          shouldObserve: (_, context) => TRACKED_OPERATIONS.includes(context?.params?.operationName)
```

because I think the signature is the same as `fillLabelsFn`.

packages/plugins/prometheus/src/utils.ts
```diff
@@ -104,23 +115,34 @@ export function registerHistogram<LabelNames extends string>(
...
+  Phases extends string[],
   LabelNames extends string,
   Params extends Record<string, any> = FillLabelsFnParams,
 >(options: {
   registry: Registry;
   histogram: Omit<HistogramConfiguration<LabelNames>, 'registers'>;
   fillLabelsFn: FillLabelsFn<LabelNames, Params>;
-}): HistogramAndLabels<LabelNames, Params> {
+  phases: Phases;
+  shouldObserve?: ShouldObservePredicate<Params>;
```
klippx, line 153: Can you ensure that if I hover this a helpful `jsdoc` will appear to explains what this option is doing, and what the default is? 

Such as
```
  /**
   * A function that determines whether the metric should be recorded or not.
   * 
   * Defaults to `() => true`.
   */
```

Might need to be repeated 2 other places below.

klippx (approved):

klippx:
I think this would cover my needs for sure, it looks great!

PR #2392, opened 2025-01-27
feat(response-cache): `onTtl` hook to manipulate the cached response

ardatan:
`onTtl` hook allows to manipulate the calculated TTL

PR #2394, opened 2025-01-27
docs(response-cache): mention `onTtl` in README

ardatan:
Mention `onTtl` hook in the readme

Fixes https://github.com/n1ru4l/envelop/pull/2392#issuecomment-2616065615

### DIFF
--- a/packages/plugins/response-cache/README.md
+++ b/packages/plugins/response-cache/README.md
@@ -695,6 +695,63 @@ const getEnveloped = envelop({
 })
 ```
 
+### Manipulate the calculated TTL
+
+If you have a some kind of custom logic, that should be used to calculate the TTL for a specific
+reason. The following example tracks the `Cache-Control` header from a remote server and uses it to
+calculate the TTL.
+
+```ts
+const ttlByContext = new WeakMap()
+
+const getEnveloped = envelop({
+  parse,
+  validate,
+  execute,
+  subscribe,
+  plugins: [
+    useSchema(
+      makeExecutableSchema({
+        typeDefs: /* GraphQL */ `
+          type Query {
+            dataFromRemote: String
+          }
+        `,
+        resolvers: {
+          Query: {
+            dataFromRemote: async (_, __, context) => {
+              const res = await fetch('https://api.example.com/data')
+              const cacheControlHeader = res.headers.get('Cache-Control')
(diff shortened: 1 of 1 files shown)


packages/plugins/response-cache/README.md
```diff
@@ -695,6 +695,63 @@ const getEnveloped = envelop({
...
+              const res = await fetch('https://api.example.com/data')
+              const cacheControlHeader = res.headers.get('Cache-Control')
+              if (cacheControlHeader) {
+                const maxAgeInSeconds = cacheControlHeader.match(/max-age=(\d+)/)
+                if (maxAgeInSeconds) {
+                  const ttl = parseInt(maxAgeInSeconds[1]) * 1000
+                  let existingTtl = ttlByContext.get(context)
+                  if (existingTtl == null || ttl < existingTtl) {
+                    ttlByContext.set(context, ttl)
+                  }
```
EmrysMyrddin: Let's not over-complicate it with WeakMaps ?
Since this is user code, it's ok to mutate the context I think. It would be clearer as an example of the TTL hook
```suggestion
                  if (context.ttl == null || ttl < context.ttl) {
                    context.ttl = ttl;
                  }
```

packages/plugins/response-cache/README.md
```diff
@@ -695,6 +695,63 @@ const getEnveloped = envelop({
...
+      })
+    ),
+    useResponseCache({
+      session: () => null,
+      onTtl({ ttl, context }) {
+        const ttlFromContext = ttlByContext.get(context)
+        if (ttlFromContext != null && ttlFromContext < ttl) {
+          return ttlFromContext
+        }
+        return ttl
```
EmrysMyrddin: Follow up of my previous comment
```suggestion
        if (context.ttl != null && context.ttl < ttl) {
          return context.ttl
        }
        return ttl
```

packages/plugins/response-cache/README.md
ardatan: Makes sense! Changed it!

packages/plugins/response-cache/README.md
ardatan: Makes sense! Changed it!

EmrysMyrddin (approved):

PR #2430, opened 2025-02-15
feat(core) New instruments API

EmrysMyrddin:
## Description

This PR is a proposal for a new API in the core.

The goal is to provide a reliable way of wrapping each phases of the GraphQL pipeline, including plugin hooks.
The main use case aimed by this new API is Observability (Monitoring and Tracing) of the GraphQL server.

### Design choices

 - When multiple instruments are found, they are composed in the same order they are defined in the plugin array. It's up to the user to handle manually compose instruments if it needs another order.
 - Instruments wrappers don't have access to phases arguments or results. Existing hooks must be used to access those data. The only goal of the tracer is to wrap the phase in a function.
 - Implied by previous point, instruments can't modify phases arguments or results.
 - The context is passed to all instruments wrappers. It is not advised (yet possible) to modify the context, but to use it as a key of weak map to communicate data with hooks.
 - The context is passed in a payload object, which will simplify the extension of this API in Yoga and Hive Gateway where not all phase have access to the graphql context.
 - It is up to the implementor of the tracer to always call the wrapped function, and to handle possible promises in case of `execute` and `subscribe` phases.

### Example of usage

```ts
import otel, type { Context } form "@opentelemetry/core"

function useOTEL(): Plugin {
  const rootSpans = new WeakMap<any, Context>();

  return {
    tracer: {
      execute: ({context}, wrapped) => otel.trace.startActiveSpan(
        'graphql.execute',
        { kind: otel.SpanKind.INTERNAL },
        rootSpans.get(context), // The new span will be a child of the span started in `onEnveloped` hook
        async () => {
          await wrapped();
          rootSpans.get(contxet).end();
        },
      );
    },
    onEnveloped({ context }) {
      const span = otel.trace.startSpan('graphql');
      rootSpans.set(context, otel.trace.setSpan(otel.context.active(), span));
    },
    onExecute: ({ context }) => ({
      // Because of the instruments wrapping the entire execution phase, including hooks,
      // we can rely on the AsyncLocalStorage used by OTEL to keep track of the current span.
      otel.trace.getSpan().setAttribute('operationName', context.operationName);
    }),
  };
}
```

## Type of change


packages/types/src/plugin.ts
```diff
@@ -43,3 +45,18 @@ export interface Plugin<PluginContext extends Record<string, any> = {}> {
    */
   onContextBuilding?: OnContextBuildingHook<PluginContext>;
 }
+
+export type Tracer<TContext extends Record<string, any>> = {
```
dotansimha: So the Tracer manages the lifecycle of the whole req? I wonder if the tracer needs more core info or that's something that's up to the end-user (or, gw) to decide in their implementation?

packages/core/src/tracer.ts
```diff
@@ -0,0 +1,60 @@
...
+        throw new Error('A plugin has already declared a tracer. Only one tracer is allowed');
+      }
+      tracer = plugin.tracer;
+    }
+  }
+  return tracer;
+}
+
+export const getTraced = <TResult, TArgs extends any[]>(context: any) => ({
+  sync(
```
dotansimha: nit: `sync` feels like `synchronize` when it's being called like that. 
maybe `traceSyncFn` or something?

packages/types/src/plugin.ts
EmrysMyrddin: My idea was to provide the very minimal info there. If you need more infos, you have the hooks.
I give access to the context (and later to the request in Yoga, and the subgraph execution request in Hive Gateway) to have a stable reference you can use as a key in a weakmap for this purpose.

For example, in OTEL, I would start the span in the tracer, and then add relevant attributes in the hooks. Hooks will have access to the span either by using OTEL context manager, or via a weakmap using the context as key.

packages/core/src/tracer.ts
EmrysMyrddin: what about `fn` and `asyncFn` ? Because you already have the `traced` part in the variable name at usage:
```ts
const traced = getTraced(context)
const tracedFn = traced.fn(fn)
const asyncTracedFn = traced.asyncFn(asyncFn)
```

packages/core/src/tracer.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!tracer) {
+      return wrapped;
+    }
+
+    return (...args) => {
+      let result: PromiseOrValue<TResult>;
+      return mapMaybePromise(
+        tracer(payload, () => {
+          result = wrapped(...args);
+          return isPromise(result) ? result.then(undefined) : undefined;
```
EmrysMyrddin:
```suggestion
          return isPromise(result) ? result.then(() =>undefined) : undefined;
```

ardatan (approved):

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
...
+import { getInstrumentsAndPlugins } from '@envelop/instruments'
+
+const { pluginInstruments, plugins } = getInstrumentsAndPlugins([
+  // put you plugin list here. This list can contain plugins with and without instruments.
+])
+
+const instruments = composeInstruments(pluginInstruments)
+
+const getEnveloped = envelop({
+  plugins: [...plugins, { instruments }]
```
jdolle: the point of this then is to be able to extract the instruments from an existing list of plugins, and then tack them on the end of the plugin list?

This is an interesting workaround and could work. An alternative approach might be to set a priority number on plugins, so that regardless of the order they are defined in, they are executed based on priority. We could then allocate certain ranges for certain things.

packages/instruments/package.json
```diff
@@ -0,0 +1,67 @@
...
+        "types": "./dist/typings/*.d.ts",
+        "default": "./dist/esm/*.js"
+      }
+    },
+    "./package.json": "./package.json"
+  },
+  "typings": "dist/typings/index.d.ts",
+  "dependencies": {
+    "@whatwg-node/promise-helpers": "^1.2.1",
+    "tslib": "^2.5.0"
```
jdolle, line 51: Isnt this usual a dev and peer dependency?

ardatan: tslib is imported and used by the JavaScript files compiled by tsc when you have "importHelpers" enabled to dedupe helper methods in the compiled codd

packages/instruments/README.md
EmrysMyrddin: That's closed to the idea yes.

We already have a defined order of plugins execution : It's the order they are defined in the array.
The problem is that, in the case of "wrapping phases", there is often the need to have a different order for "wrapping" and for normal hooks.

So the point of those utils are to extract instruments from plugins (and remove them), so that you can compose them is the order you want. Since you know there is no other instruments, you can put it in the end of the plugin, or at the beginning, it doesn't matter :-)

packages/instruments/package.json
EmrysMyrddin: I don't know to be honest, it is like this in other packages so I just copied the same setup ^^'

.github/workflows/tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install Dependencies using pnpm
         run: pnpm install --no-frozen-lockfile && git checkout pnpm-lock.yaml
       - name: Build
-        run: pnpm run ts:check
+        run: pnpm run build
```
enisdenjo, line 36: The name is wrong, not the command.
```suggestion
      - name: Check TypeScript
        run: pnpm run ts:check
```

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
+## `@envelop/instruments`
+
+This package contains uitility functions and types to ease the use of instruments accross Envelop,
+Yoga, wathwg-node and Hive Gateway plugins.
```
enisdenjo:
```suggestion
This package contains utility functions and types to ease the use of instruments accross Envelop,
Yoga, whatwg-node and Hive Gateway plugins.
```

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### `getInstrumentsAndPlugins(plugins: Plugin[]): { pluginInstruments: Instruments[], plugins: Plugin[] }`
+
+This function extracts the instruments from the plugins and returns both the extracted instruments
+and the plugins without their `instruments` field.
+
+This is usefull when you want to customize the execution order of the instruments.
```
enisdenjo:
```suggestion
This is useful when you want to customize the execution order of the instruments.
```

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+const { pluginInstruments, plugins } = getInstrumentsAndPlugins([
+  // put you plugin list here. This list can contain plugins with and without instruments.
+])
+```
+
+## `composeInstruments(instruments: Instruments[]): Instruments`
+
+This function composes all the instruments into one. The instruments will be called in the same
+order than they are in the array (first is outter most call, last is inner most).
```
enisdenjo:
```suggestion
order than they are in the array (first is outer most call, last is inner most).
```

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This can be used in conjonction with `getInstrumentsAndPlugins` function to customize the order of
```
enisdenjo:
```suggestion
This can be used in conjunction with `getInstrumentsAndPlugins` function to customize the order of
```

packages/instruments/README.md
```diff
@@ -0,0 +1,41 @@
...
+execution of the instruments if the default one doesn't suites your needs.
```
enisdenjo:
```suggestion
execution of the instruments if the default one doesn't suite your needs.
```

.changeset/tough-ears-suffer.md
```diff
@@ -0,0 +1,7 @@
+---
+'@envelop/types': minor
+'@envelop/core': minor
+'@envelop/instruments': major
+---
+
+Add new Tracer API
```
enisdenjo: It's called "instruments" now, right?

Also, please add some more information and an example to the changelog - lets increase awareness.

enisdenjo (changes requested):
Please also add tests for chaining without using the helper utilities (the default order of plugins).

Additionally, I'd this information somewhere in the documentation - people need to use the helpers ONLY if they want custom order of instrumentation.

enisdenjo:
Please also update the description of this PR to reflect your latest changes. People will land here and we dont want misleading information.

.github/workflows/tests.yml
EmrysMyrddin: We need to actually build the package for `pnpm bob check` to actually check the correctness of the ESM built version

ardatan: https://guild-oss.slack.com/archives/CSFK5GGAU/p1740658884058789?thread_ts=1740658470.220519&cid=CSFK5GGAU

enisdenjo: So `ts:check` is unused?

enisdenjo:
[Description still uses the `tracer` in the example.](https://github.com/n1ru4l/envelop/pull/2430#issue-2855499803)

.changeset/tough-ears-suffer.md
```diff
@@ -0,0 +1,92 @@
...
+This allows to have a different execution order of hooks and instruments.
+
+
+```ts
+import { useSentry } from '@envelop/sentry'
+import { useOpentelemetry } from '@envelop/opentelemetry'
+
+const { instruments: sentryInstruments, ...sentryPlugin } = useSentry();
+const { instruments: otelInstruments, ...otelPlugin } = useOpentelemetry();
+const instruments = composeInstruments([otelInstruments, sentryInstruments])
```
enisdenjo, line 75: `composeInstruments` is nowhere imported.

packages/instruments/README.md
```diff
@@ -0,0 +1,30 @@
...
+> ordering doesn't suite your need (ie. instruments and hooks should not be executed in the same
+> order)
+
+## `composeInstruments(instruments: Instruments[]): Instruments`
+
+This function composes all the instruments into one. The instruments will be called in the same
+order than they are in the array (first is outer most call, last is inner most).
+
+```ts
+import { getInstrumentsAndPlugins } from '@envelop/instruments'
```
enisdenjo: This import is not used. Did you mean:
```suggestion
import { composeInstruments } from '@envelop/instruments'
```

or

```suggestion
import { composeInstruments } from '@envelop/core'
```

if it's coming from core?

packages/instruments/README.md
```diff
@@ -0,0 +1,30 @@
+## `@envelop/instruments`
+
+This package contains utility functions and types to ease the use of instruments accross Envelop,
+Yoga, whatwg-node and Hive Gateway plugins.
+
+> [!NOTE] Instruments are automatically composed together. This should only be used if the default
+> ordering doesn't suite your need (ie. instruments and hooks should not be executed in the same
```
enisdenjo: Typos and less negation.
```suggestion
> ordering doesn't suit your needs (ie. instruments and hooks should be executed in different
```

packages/instruments/README.md
```diff
@@ -0,0 +1,30 @@
...
+> order)
+
+## `composeInstruments(instruments: Instruments[]): Instruments`
+
+This function composes all the instruments into one. The instruments will be called in the same
+order than they are in the array (first is outer most call, last is inner most).
```
enisdenjo:
```suggestion
order as they are in the array (from top to bottom).
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
 - `schema` - the `GraphQLSchema`
 - `replaceSchema` - replaces the schema. Calling this will trigger `onSchemaChange` for all other
   plugins (except for the one that initiated the change);
+
+### `instruments`
+
+An optional `instruments` instance can be present in the plugin.
+
+This `Instruments` instance allows to wrap an entire phase execution (including all plugin hooks),
```
enisdenjo:
```suggestion
This `Instruments` instance allows wrapping an entire phase execution (including all plugin hooks);
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+meaning running code just before, just after and around the execution of the phase.
```
enisdenjo:
```suggestion
meaning, running code just before, just after and around the execution of the phase.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+Instruments doesn't have access to input/output of a phase, use hooks to have access to those data.
```
enisdenjo:
```suggestion
Instruments don't have access to the input and outputs of a phase, use hooks to access that data.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+If needed, we recommend to share data between instruments and hooks with a `WeakMap` and the given
+`context` as the key.
+
+All instruments takes 2 parameters:
```
enisdenjo:
```suggestion
All instruments take 2 parameters:
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+- `payload`: an object containing the graphql context.
+- `wrapped`: The function representing the execution of the phase. It takes no parameters, and
+  returns `void` (or `Promise<void>` for asynchrone phases). **This function must always be
```
enisdenjo:
```suggestion
  returns `void` (or `Promise<void>` for asynchronous phases). **This function must always be
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+  called**.
+
+#### Instruments composition
+
+If multiple plugins have `instruments`, they are composed in the same order they are defined the
```
enisdenjo:
```suggestion
If multiple plugins have `instruments`, they are composed in the same order as they are defined in the
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+plugin array (the first is outtermost call, the last is inner most call).
```
enisdenjo:
```suggestion
plugin array (from top to bottom).
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+It is possible to customize this composition if it doesn't suite your need (ie. you need hooks and
```
enisdenjo:
```suggestion
It is possible to customize this composition if it doesn't suit your needs (ie. you need hooks and
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+instruments to have a different oreder of execution).
```
enisdenjo:
```suggestion
instruments to have a different order of execution).
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+})
+```
+
+#### `init`
+
+Wraps the envelop (the call to `envelop` function) initialisation.
+
+This includes all the plugins `onEnveloped` hooks, and the creation of the Envelop Orchestrator.
+
+This instruments must be synchrone, the wrapped function is always synchrone.
```
enisdenjo:
```suggestion
This instrument must be synchronous, the wrapped function is always synchronous.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+#### `parse`
+
+Wraps the parse phase. This includes all the plugins `onParse` hooks and the actual engine `parse`.
```
enisdenjo: "engine" might be confusing.
```suggestion
Wraps the parse phase. This includes all the plugins `onParse` hooks and the actual `parse` function.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+This instrument must be synchrone, the wrapped function is always synchrone.
```
enisdenjo:
```suggestion
This instrument must be synchronous, the wrapped function is always synchronous.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+#### `validate`
+
+Wraps the validate phase. This includes all the plugins `onValidate` hooks and the actual engine
+`validate`.
```
enisdenjo:
```suggestion
Wraps the validate phase. This includes all the plugins `onValidate` hooks and the actual `validate`
function.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+This instrument must be synchrone, the wrapped function is always synchrone.
```
enisdenjo:
```suggestion
This instrument must be synchronous, the wrapped function is always synchronous.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+#### `context`
+
+Wraps the context building phase. This includes all the plugins `onContextBuilding` hooks.
+
+This instrument must be synchrone, the wrapped function is always synchrone.
```
enisdenjo:
```suggestion
This instrument must be synchronous, the wrapped function is always synchronous.
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+
+#### `execute`
+
+Wraps the execute phase. This includes all the plugins `onExecute` hooks.
+
+This instrument can be asynchrone, the wrapped function **can be** asynchrone. Be sure to `await` or
```
enisdenjo:
```suggestion
This instrument can be asynchronous, the wrapped function **can be** asynchronous. Be sure to `await` or
```

website/src/pages/docs/plugins/lifecycle.mdx
```diff
@@ -267,3 +267,88 @@ Some plugins (like gateway implementations) could potentially change the schema
...
+use `.then` on the result of the `wrapped` function to run code after the `execute` phase.
+
+Note that `wrapped` is not guaranted to return a promise.
+
+#### `subscribe`
+
+Wraps the subscribe phase. This includes all the plugins `onSubsribe` hooks. Note that it doesn't
+wrap the entire lifetime of the subscription, but only it's intialisation.
+
+This instrument can be asynchrone, the wrapped function **can be** asynchrone. Be sure to `await` or
```
enisdenjo:
```suggestion
This instrument can be asynchronous, the wrapped function **can be** asynchronous. Be sure to `await` or
```

enisdenjo (changes requested):
I don't see tests where `composeInstruments` is not used. Please add tests that make sure the order of plugins is kept if instruments are not composed:

```ts
import { useSentry } from '@envelop/sentry'
import { useOpentelemetry } from '@envelop/opentelemetry'

  plugins: [useSentry(), useOpentelemetry() ],
})
```

.github/workflows/tests.yml
EmrysMyrddin: I suppose yes, since the build is already checking types too.

I use it locally while coding.

packages/core/test/instruments.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+            return { test: 'foo' };
+          },
+          validate: () => {
+            result.push('validate');
+            return 'test';
+          },
+        }),
+        make('1'),
+        make('2'),
+        make('3'),
```
EmrysMyrddin, line 63: @enisdenjo Here the plugins are used as is, without manual composition.

You can see in the expect later that I'm checking that the order is correct and reflect the same order than the array.

packages/instruments/README.md
EmrysMyrddin: Yeah it's a leftover, I thought it would be useful but in fact it's not.

It is coming from core because core re-exports `@envelop/instruments`. Is though it would be easier like this, since the core already depends on it.

enisdenjo (approved):

PR #2665, opened 2025-08-16
fix(withState): respect getters and setters of non hook properties

EmrysMyrddin:
The `withState` plugin utility was not respecting plugin properties with getters or setters.

In particular, properties with getters were just copied at plugin creating time, leading to potential bugs where the getter doesn't always return the same thing over time.

This PR fixes this by enumerating all properties with `Object.getOwnPropertyDescriptors` and iterating over it instead of `Object.entries`. This also allows to keep the `enumerable`, `writable` and `configurable` flags from each non-hooks properties.

See this PR for a usage of this: https://github.com/graphql-hive/gateway/pull/1360

### DIFF
--- a/.changeset/plain-trams-kneel.md
+++ b/.changeset/plain-trams-kneel.md
@@ -1,0 +1,6 @@
+---
+'@envelop/core': patch
+---
+
+Fix getters and setters being stripped away. The value was copied at plugin creation instead of
+copying the getter and setter (if any).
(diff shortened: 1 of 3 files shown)


enisdenjo (approved):

ardatan (changes requested):
Do you think failing tests are related?

EmrysMyrddin:
Weird, no I don't see how it could be linked :/ the test doesn't even use `withState`

EmrysMyrddin:
@ardatan Test are passing again after a rebase :-)

ardatan (approved):

PR #2676, opened 2025-09-11
feat(rate-limiter): improve performance

EmrysMyrddin:
## Description

To apply the configuration based on type and field name, the rate limiter plugin is using `minimatch` to allow glob patterns and simplify the configuration.

But `minimatch` implementation is very unefficient. It uses `RegExp` under the hood (which is already pretty expensive), but it doesn't even memoize the `RegExp` instances. This means that a new one is created and compiled for each resolver, for each configured type/field.

Having such heavy process in a hot path like this one can have a big impact of performance.

This PR replaces `minimatch` by `picomatch` and builds all relevant matchers ahead of time. This way, only the actual string test is done in the hot path.

It also moves all the logic for replacing the resolver based on configuration to the `onSchemaChange` hook, to minimize the overhead at execution time.

Realted to https://github.com/graphql-hive/gateway/issues/1485

## Type of change

## Further comments

By reworking the plugin, I also came across some small issues:
 - The `directiveName` option was not honored.
 - Directives on fields with a matching configuration was silently ignored
 - Configurations matching the same field was silently ignored (only the first matching config was used)

### DIFF
--- a/.changeset/@envelop_rate-limiter-2676-dependencies.md
+++ b/.changeset/@envelop_rate-limiter-2676-dependencies.md
@@ -1,0 +1,7 @@
+---
+"@envelop/rate-limiter": patch
+---
+dependencies updates:
+  - Added dependency [`@types/picomatch@^4.0.2` ↗︎](https://www.npmjs.com/package/@types/picomatch/v/4.0.2) (to `dependencies`)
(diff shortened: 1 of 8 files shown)


ardatan (approved):
You can also use mapSchema but this is also good

EmrysMyrddin:
Yes, that was my first thought, and then I found out that `useOnResolve` is not using mapSchema but mutating it in place, so I've done the same :-)

PR #2715, opened 2025-10-02
chore(deps): support sentry v10

thomasleduc:
the pull request._

## Description

Update sentry plugin to be used with sentry v10 (before sentry v8).

#### Since Sentry SDK v10, you don’t need to import '@sentry/tracing' anymore.

Remove the imports in documentation and test, already embedded in sentry/node or sentry/nextjs.

#### sentry/types

Remove sentry/types (deprecated) and install sentry/core where the types can be found now.

## Type of change

      expected)

There is a breaking change since sentry is in peerDependencies and need to be updated to 10.

## How Has This Been Tested?

The tests didn't changed and cover the same amount of code. 

**Test Environment**:

- OS: Mac 15.6.1
- `@envelop/sentry`:
- NodeJS: v22.13.1

## Checklist:

- [x] I have followed the

## Further comments

There is a discussion to address on whether we do a sentry/node | sentry/nextjs version of envelop/sentry plugin or not.
Now that sentry is embedding a lot in those imports, users may want to have only sentry/nextjs or sentry/node that do basically the same thing.
I choose to keep sentry/node for this plugin as it was before.

### DIFF
--- a/packages/plugins/sentry/README.md
+++ b/packages/plugins/sentry/README.md
@@ -37,8 +37,6 @@ yarn add @sentry/node @sentry/tracing @envelop/sentry
 import { execute, parse, specifiedRules, subscribe, validate } from 'graphql'
 import { envelop, useEngine } from '@envelop/core'
 import { useSentry } from '@envelop/sentry'
-// do this only once in you entry file.
-import '@sentry/tracing'
 
 const getEnveloped = envelop({
   plugins: [
--- a/packages/plugins/sentry/package.json
+++ b/packages/plugins/sentry/package.json
@@ -48,8 +48,8 @@
   },
   "peerDependencies": {
     "@envelop/core": "workspace:^",
-    "@sentry/core": "^10.17.0",
-    "@sentry/node": "^10.17.0",
+    "@sentry/core": "^8.0.0 || ^9.0.0 || ^10.0.0",
+    "@sentry/node": "^8.0.0 || ^9.0.0 || ^10.0.0",
     "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
--- a/packages/plugins/sentry/package.json
+++ b/packages/plugins/sentry/package.json
@@ -48,7 +48,8 @@
   },
   "peerDependencies": {
     "@envelop/core": "workspace:^",
-    "@sentry/node": "^8.0.0",
+    "@sentry/core": "^10.17.0",
+    "@sentry/node": "^10.17.0",
     "graphql": "^14.0.0 || ^15.0.0 || ^16.0.0"
   },
   "dependencies": {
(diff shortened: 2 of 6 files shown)


packages/plugins/sentry/README.md
```diff
@@ -37,8 +37,6 @@ yarn add @sentry/node @sentry/tracing @envelop/sentry
 import { execute, parse, specifiedRules, subscribe, validate } from 'graphql'
 import { envelop, useEngine } from '@envelop/core'
 import { useSentry } from '@envelop/sentry'
-// do this only once in you entry file.
-import '@sentry/tracing'
```
ardatan, line 41: Is this not needed anymore?

ardatan:
Thanks for the PR! Could you also add a changeset?

packages/plugins/sentry/README.md
thomasleduc: Nope, as I wrote in the PR description : 
>Since Sentry SDK v10, you don’t need to import '@sentry/tracing' anymore.
> Removed the imports in documentation and test, already embedded in sentry/node or sentry/nextjs.

I can try to retrieve some sentry migration link if you want.

thomasleduc: https://github.com/getsentry/sentry-javascript/issues/5815#issuecomment-1480936427

packages/plugins/sentry/package.json
```diff
@@ -48,7 +48,8 @@
   },
   "peerDependencies": {
     "@envelop/core": "workspace:^",
-    "@sentry/node": "^8.0.0",
+    "@sentry/core": "^10.17.0",
```
ardatan: I think the code is still compatible with v8 and v9 right? So we can keep v8 and v9 in peerDependencies?

thomasleduc: I don't know we removed global import of sentry/tracing which is now embedded in sentry/node. the type also are not compatible even if runtimewise it's not a problem. 
Not sure we can consider it as compatible.

ardatan: If so, let's make it a major release then because it removes the compatibility of older Sentry versions. [chore(deps): sentry: Update sentry to ^10.17 by thomasleduc · Pull Request #2715 · graphql-hive/envelop](https://github.com/graphql-hive/envelop/pull/2715/files#diff-ed4497e7ec0dbe27ac5f4d994820536e711244628618d4705164d59107a2aa6cR2)

thomasleduc: Yes you're right. 
I have another thing to check. A parameter not tested yet in the plugin and that I don't use seems broken.

thomasleduc: TraceparentData is the old way to use Sentry.
Should I kept the original option name (`tranceparentData`) for minimal change, even if I change the shape for
something like `{ sentryTrace?: string; baggage?: string }` instead of the old TraceparentData ?

ardatan (approved):

PR #2815, opened 2025-12-17
feat(response-cache): shared key for public responses

EmrysMyrddin:
This is a re-write based on the work of @k-i-k-s PR: #2202

It solves 2 issues from the initial implementation:
 - On schema reload, most of the stale config per schema coordinate is kept, potentially leading to data leaks
 - The scope calculation based on the document only can't reliably guess the scope when Union or Interface types are involved.

The design choice in this PR is to narrow the flexibility of the original proposition but to address more directly the original need: automatically cache public responses under a shared key among all users.

The choice I've made is that the key can be built a second time (only if the option is enabled) at cache write time, based on the computed response scope. This way, it doesn't add any overhead when disabled and reduces the added complexity of this feature.


packages/plugins/response-cache/src/plugin.ts
```diff
@@ -572,28 +649,34 @@ export function useResponseCache<PluginContext extends Record<string, any> = {}>
             console.warn(
               '[useResponseCache] Cache instance is not available for the context. Skipping cache lookup.',
             );
+            return;
```
EmrysMyrddin, line 652: Add a test for this, it was clearly a bug, a return statement was missing. No idea why TS is not catching this bug...

packages/plugins/response-cache/test/response-cache.spec.ts
```diff
@@ -3524,101 +3524,7 @@ describe('useResponseCache', () => {
...
-        `;
-
-        await testInstance.execute(query);
-
-        expect(spy).toHaveBeenCalledTimes(1);
-      });
-    });
-
-    it('should return PRIVATE scope in buildResponseCacheKey even when requesting property from a fragment', async () => {
+    it.only('should ignore session id for responses with public key', async () => {
```
k-i-k-s, line 3527: Is it intended to keep `only` ?

PR #2852, opened 2026-01-25
fix(opentelemetry): activate resolver span context

kdawgwilk:
## Summary
- wrap resolver execution in the resolver span context so nested spans attach correctly
- add coverage to assert nested spans are parented by resolver spans

## Test plan
- [ ] Not run (not requested)


kdawgwilk:
i tried deploying this to production with a `patch-package` and it caused a call stack exceeded errors so its not ready to go yet

PR #2868, opened 2026-02-11
fix: use context extension context for the error handler

n1ru4l:
## Description

Some properties that are in my context are not accessible in the error handler if an error is raised within a further context building hook.

## Type of change

## How Has This Been Tested?

- [x] Unit test

### DIFF
--- a/packages/core/src/plugins/use-error-handler.ts
+++ b/packages/core/src/plugins/use-error-handler.ts
@@ -63,8 +63,8 @@ export const useErrorHandler = <ContextType extends Record<string, any>>(
         }
       };
     },
-    onPluginInit(context) {
-      context.registerContextErrorHandler(({ error, context }) => {
+    onPluginInit({ registerContextErrorHandler }) {
+      registerContextErrorHandler(({ error, context }) => {
         if (isGraphQLError(error)) {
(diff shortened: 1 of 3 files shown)


packages/core/src/plugins/use-error-handler.ts
```diff
@@ -64,7 +64,7 @@ export const useErrorHandler = <ContextType extends Record<string, any>>(
       };
     },
     onPluginInit(context) {
```
ardatan: It is weird that this argument is called context because it is actually payload 😅

ardatan:
```suggestion
    onPluginInit({ registerContextErrorHandler }) {
```

packages/core/src/plugins/use-error-handler.ts
```diff
@@ -64,7 +64,7 @@ export const useErrorHandler = <ContextType extends Record<string, any>>(
...
-      context.registerContextErrorHandler(({ error }) => {
+      context.registerContextErrorHandler(({ error, context }) => {
```
ardatan:
```suggestion
      registerContextErrorHandler(({ error, context }) => {
```

ardatan (approved):

dotansimha (approved):

